PR #24, opened 2020-06-30
feat: externalIgnore

uneco:
closes #23

PR #26, opened 2020-07-01
feat: exclude option

### LINKED ISSUE #23: feature request: Option to selection of packages to be excluded from externals
Besides the 'aws-sdk' hard-coded in [this line](https://github.com/floydspace/serverless-esbuild/blob/master/src/packExternalModules.ts#L145), there are other libraries, such as the one provided by Lambda Layers (often for faster builds).

uneco:
closes #23 - This approach makes more sense than #24.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,8 @@ custom:
     minify: false
 ```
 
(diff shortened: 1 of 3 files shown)


floydspace:
Hey @uneco . Thank you for your effort, yes it does make sense

PR #50, opened 2020-07-24
Improve Example

Hebilicious:
Use async return style in the handler


Hebilicious:
@floydspace 
I added an integration tests and a basic CI workflow in the last 2 commits, have a look

Hebilicious:
@floydspace CI ran properly in my fork : https://github.com/Hebilicious/serverless-esbuild/runs/911889490?check_suite_focus=true

floydspace:
Hi @Hebilicious , thank you for your contribution, could you please explain why to overwhelm the example code, it's supposed to be like minimal working example of using `serverless-esbuild`, not sure should we make it more complicated. Thou integration test could cover the fact that the plugin works

Hebilicious:
I was working on my repo https://github.com/Hebilicious/serverless-esbuild-template and used your example at the starting point, therefore I wanted to contribute here as well.
The handler code is the same, as you said the added complexity is just here for the test coverage for the actual plugin.
Maybe we could have 2 examples if you want to keep a very minimal one as well.

PR #87, opened 2021-02-22
feat(package): optimize bundle for package individually

olup:
This PR allows for some optimization on the individual packaging. This is a WIP to get feedback from the main maintainer.


olup:
@floydspace I'd like your input on this - the flow is here but code could - should - probably be optimized.

floydspace:
@olup thank you very much for the contribution, really appreciate, will check it during the day 👍

olup:
Actually, I am not too happy with this implementation. It works, but could be better. I think I can skip the copy / install / prune all together by only getting the list of required deps from the packager (from the intersection of externals and actually required dependency in the bundle) - and from there build the zip file by selecting what I allow inside. This would be very neat, I'll update the pr once I find a way to do that.

floydspace:
@olup to be honest, I tested it and it didn't work for me. I had an issue with copying. Yes please take your time, it would be nice to have the fix. thank you

olup:
I just hard-pushed the branch with a way better system.
It works very well on one of my company's prod serverless repo, but more testing should be done. Some optimization is still required too. But this is a start.

floydspace:
Hey @olup I tested, it worked perfectly for the case when we need to package externals.
Then I tested the case when we bundle everything, here I got an error `Error: Cannot find module '/Users/victor/Projects/_own/individually/.build/package.json'`, so I fixed it by replacing `buildDir` with `findProjectRoot()` in line 54, so it works again
And also I fixed a bug with `yarn` packager.
Could you please check my commit from your perspective? Really appreciate your efforts. let me know if everything's fine I will merge the feature.

olup:
Ok, I saw the commit. Thank you for fixing the yarn bug ! However on the other thing : the no external case is indeed missing in my code, but I think we need to read the generated package.json, not the original, because it contains only references to the external, and have been generated after some logic already. 
What we are doing here is to get the list of the externals of the complete project (in the generated package json) and reduce this list to only what appears in the actual bundle. So using the original package json will work too I guess, but I think it would be better to check for existence of the generated package.jdon, or to simply check if there are externals declared at all - and skip the whole part when it's not the case. I'll commit something in the morning and we can test, if you agree with what I said - but if I am being mistaken, please tell me.

floydspace:
@olup yes go ahead, I agree, this logic should be skipped if we do not use externals

olup:
Just pushed. This makes the external part dependent on if there is externals or not. I also used the declaration of the externals as my comparison array. I wonder if it's ok (I mean, no use of regex and all in this, right ?).

This works with my test repo, with and without externals.

Also added bundle size in the logs, just because it's nice :-)

floydspace:
@olup it works perfectly on my test case. Merging it

PR #92, opened 2021-03-01
fix(package): exclude complete bundle directory

olup:
This PR excludes the other handlers' bundles directory (as opposed to bundler file before), thus ignoring also source maps in the individual packaging.


floydspace (approved):
looks good

mitchh:
I'm having trouble with this where I have

PR #95, opened 2021-03-06
fix: don't include siblings in folder when packaging individually

mitchh:
#92 helped keep from including extra map files, but it still includes too much. 

Storing `dirname` erases too much information and we cant distinguish siblings. So given:

```
/folder1/
/folder1/foo.ts
/folder1/bar.ts
[...]
/folder2/
/folder2/bat.ts
[...]
```

foo.zip gets both foo and bar (and their maps), and so does bar.zip, and so on.

This change assumes that there may be generated files that are not listed in `buildResults`, but that they are "prefix-named", that is `foo.js` => `foo.js.map` - generalizing, we ignore here `foo.js*`

### DIFF
--- a/src/pack.ts
+++ b/src/pack.ts
@@ -70,7 +70,7 @@ export async function pack(this: EsbuildPlugin) {
 
   // get a list of every function bundle
   const buildResults = this.buildResults;
-  const bundlePathList = buildResults.map(b => path.dirname(b.bundlePath));
+  const bundlePathList = buildResults.map(b => b.bundlePath);
 
   // get a list of externals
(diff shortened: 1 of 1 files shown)


olup:
Sounds good !

olup:
Any reason why this is still a draft ?

olup:
@floydspace I am merging it ad this is pretty basic, but very much needed changes.

mitchh:
Thanks for the quick merge @olup 

> Any reason why this is still a draft ?

Just that I did it late at night and was going to revisit it in the morning and see if it needed any polish :-).  Thanks again.

PR #101, opened 2021-03-16
Add esbuild cache

thomaschaaf:
This improves build times for larger projects. Sadly it does not improve the memory usage as anticipated.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -51,7 +51,7 @@ export class EsbuildPlugin implements Plugin {
 
   serverless: Serverless;
   options: OptionsExtended;
-  esbuildCache: any;
+  esbuildCache: Record<string, BuildResult>;
   hooks: Plugin.Hooks;
   buildOptions: Configuration;
   buildResults: {
@@ -163,7 +163,7 @@ export class EsbuildPlugin implements Plugin {
     chokidar.watch(this.buildOptions.watch.pattern, options).on('all', () => {
       this.bundle(true)
         .then(() => this.serverless.cli.log('Watching files for changes...'))
-        .catch(e =>
+        .catch(() =>
           this.serverless.cli.log('Bundle error, waiting for a file change to reload...')
         );
     });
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -152,17 +154,19 @@ export class EsbuildPlugin implements Plugin {
   async watch(): Promise<void> {
     const options = {
       ignored: this.buildOptions.watch.ignore,
-      awaitWriteFinish: true,
+      awaitWriteFinish: {
+        stabilityThreshold: 50,
```
thomaschaaf, line 158: this it there because before it would wait 2000 ms just until it would start compiling changes

olup: that sounds reasonable when taking into account the size of a typical code file. But if half saved code gets compiled we'll have to higher this value a little bit (I think a couple of hundreds should still give a smooth experience but prevent any weirdness)

olup:
Sounds good to me, however one question, shouldn't we `dispose` the cached build on process end ? Or is the process cleaned out anyway ?

floydspace:
@olup I had the same question in mind.
@thomaschaaf Also, I see that incremental build is running in `watch` function which is run on `before:offline:start`, but also the normal building with `incremental: false` is run as well at the beginning of the hook.
Can we somehow improve it, so the first build in the `before:offline:start` would use `incremental: true` and the watch would use `rebuild` from cache?

JoaquinFernandez:
> This improves build times for larger projects. Sadly it does not improve the memory usage as anticipated.

@thomaschaaf I investigated on how to limit memory usage, but that's on esbuild and not possible as far as I can tell. However I just submitted a PR to limit the number of lambdas that are concurrently build, so you can keep memory consumption at reasonable levels while still building amazingly fast, here is the PR in case it is helpful:

https://github.com/floydspace/serverless-esbuild/pull/201

This PR and mine would probably conflict, happy to take a look when this gets merged to update mine @floydspace

Issue #110, opened 2021-04-09
package.patterns seems to be ignored


kabo:
The README says

> All files from package/include will be included in the final build file. See [Exclude/Include](https://serverless.com/framework/docs/providers/aws/guide/packaging#exclude--include)

But package.include/exclude has been deprecated and replaced by package.patterns.

However, if I do something like

```yaml
  myFn:
    package:
      patterns:
        - resources/cert.cnf
```

that file is not included in the zip.

If I do

```yaml
  myFn:
    package:
      patterns:
        - '!**'
```

that doesn't have any effect either. It seems to just be ignored.

kabo:
Looking at https://github.com/floydspace/serverless-esbuild/blob/master/src/index.ts#L220 it appears only `include` is honoured, and only in the global `package` section, not on the individual functions.

kabo:
Ah, but `@types/serverless` hasn't been updated to support patterns... https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/serverless/index.d.ts#L57-L62

kabo:
I've made a PR to the types to support patterns
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52301

PR #111, opened 2021-04-09
fix: #110 package.patterns

kabo:
- Update to latest @types/serverless which supports package.patterns
- Updated README to mention patterns and updated link to serverless docs
- Uses package.patterns instead of patterns.include in the global package section
- Added support for package.patterns on individual functions. It does this by copying the individual files into a special folder in the build folder, which gets ignored by all other functions. Then that special folder prefix gets removed when the intended function is packaged.

I'm not sure it's the best way to solve but at least it's a way to solve it :)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -173,7 +173,6 @@ export class EsbuildPlugin implements Plugin {
       const fn = this.serverless.service.getFunction(fnName);
       fn.package = fn.package || {
         patterns: [],
-        exclude: [],
       };
 
       // Add plugin to excluded packages or an empty array if exclude is undefined
@@ -244,13 +243,10 @@ export class EsbuildPlugin implements Plugin {
     // include any "extras" from the individual function "patterns" section
     for (const fnName in this.functions) {
       const fn = this.serverless.service.getFunction(fnName);
-      const fnPatterns = [
-        ...new Set([ ...(fn.package.include || []), ...(fn.package.patterns || []) ]),
-      ];
-      if (fnPatterns.length === 0) {
+      if (fn.package.patterns.length === 0) {
         continue;
       }
-      const files = await globby(fnPatterns);
+      const files = await globby(fn.package.patterns);
       for (const filename of files) {
         const destFileName = path.resolve(path.join(this.buildDirPath, `__only_${fn.name}`, filename));
         const dirname = path.dirname(destFileName);
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,7 @@ import { build, BuildResult, BuildOptions } from 'esbuild';
 import * as fs from 'fs-extra';
 import * as globby from 'globby';
 import * as path from 'path';
-import { mergeRight } from 'ramda';
+import { concat, mergeRight } from 'ramda';
 import * as Serverless from 'serverless';
 import * as Plugin from 'serverless/classes/Plugin';
 import * as chokidar from 'chokidar';
@@ -173,11 +173,12 @@ export class EsbuildPlugin implements Plugin {
       const fn = this.serverless.service.getFunction(fnName);
       fn.package = fn.package || {
         patterns: [],
+        exclude: [],
       };
 
       // Add plugin to excluded packages or an empty array if exclude is undefined
       fn.package.patterns = [
-        ...new Set([ ...(fn.package.patterns || []), '!node_modules/serverless-esbuild' ]),
+        ...new Set([ ...(fn.package.exclude || []).map(concat('!')), ...(fn.package.patterns || []), '!node_modules/serverless-esbuild' ]),
       ];
     }
   }
@@ -220,8 +221,11 @@ export class EsbuildPlugin implements Plugin {
     const { service } = this.serverless;
 
     // include any "extras" from the "patterns" section
-    if (service.package.patterns && service.package.patterns.length > 0) {
-      const files = await globby(service.package.patterns);
+    const globalPatterns = [
+      ...new Set([ ...(service.package.include || []), ...(service.package.patterns || []) ]),
+    ];
+    if (globalPatterns.length > 0) {
+      const files = await globby(globalPatterns);
 
       for (const filename of files) {
         const destFileName = path.resolve(path.join(this.buildDirPath, filename));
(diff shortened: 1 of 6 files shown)


olup:
Hi @kabo and thank you for this pull request. I will check it later today or in the next days, but on top of my head I think you disabled include/exclude support, however we cannot assume what version of serverless people are using, especially since deprecation dates from last month. This would be a breaking change for serverless-esbuild. So I think first thing would be to support both `include/exclude` and `patterns` options.

kabo:
Cool, yeah if the approach looks good otherwise I can look into that.

olup:
Cool. There is some other stuff that will need addressing, in how the flow works and file get copied. But I don't have time just right now to check that up, so I'll comment once I've done a deeper review, especially concerning edge cases.

olup:
Hey, so I went a bit deeper on the review. Seems to me that there's still a couple of pieces missing: in global packaging, the patterns of each function are simply ignored. Also, https://github.com/floydspace/serverless-esbuild/blob/353ecd5a6d3840af62b97997405fcc66703cce41/src/pack-externals.ts#L179 uses the `exclude` props this has not been updated.

More generally I am not a huge fan of the prefixed directory. I mean, it could work like that (once addressed the global packaging issue) but i'd rather have some more "data" driven approach. It's just a sketch in my head now, but we might be better off copying all the *included* files from global and functions to the build directory, and then use the complete pattern list (inclusion and exclusion) when copying to the zip file? Or only work with includes and treat excludes only in the dependencies level, which is what I think we are doing now? Am I making enough sense?

In this view, it might be better to distribute the patterns to include and exclude rather than the contrary? But this is still rough thinking.

On a side note @floydspace, I see that we use `glob` and `globby`. We might just stick to `globby` everywhere? that would save us a `stat` directory lookup in the packing phase (because globby have an option to output files only) and could speedup build as `globby` uses `fast-glob` which *sounds* faster than glob (am I falling for the name here ?)

kabo:
> in global packaging, the patterns of each function are simply ignored
hmm, yeah, didn't think of that scenario.

But `buildOptions.exclude` is correct right? serverless has service.package.patterns and function.package.patterns, buildOptions seems to be populated from custom.esbuild, which to my understanding hasn't changed. From the README:

> The package specified in the `exclude` option is passed to esbuild as `external`, but it is not included in the function bundle either.

I don't see a reason to mess with that.

kabo:
Regarding include/exclude vs patterns my thinking was the include/exclude pattern has been deprecated in favour of `patterns`. This suggests to me that the future proof way of doing it is `patterns`, otherwise in the future we'll be stuck with our own include/exclude pattern that isn't being used anywhere else, it is then a custom pattern that maps over the pattern serverless is using and necessitates converting between the two. Sure, if we still want to support include/exclude we'll have to convert that to `patterns`, but eventually we should be able drop include/exclude and just remove that code. This was my reasoning.

kabo:
I have tested this code with:

- individual patterns, individual packaging
- individual patterns, no individual packaging
- global patterns, no individual packaging
- global patterns, individual packaging

olup:
> But `buildOptions.exclude` is correct right? serverless has service.package.patterns and function.package.patterns, buildOptions seems to be populated from custom.esbuild, which to my understanding hasn't changed. From the README:
> 
> > The package specified in the `exclude` option is passed to esbuild as `external`, but it is not included in the function bundle either.
Yes actually you are right

olup:
> Regarding include/exclude vs patterns my thinking was the include/exclude pattern has been deprecated in favour of `patterns`. This suggests to me that the future proof way of doing it is `patterns`, otherwise in the future we'll be stuck with our own include/exclude pattern that isn't being used anywhere else, it is then a custom pattern that maps over the pattern serverless is using and necessitates converting between the two. Sure, if we still want to support include/exclude we'll have to convert that to `patterns`, but eventually we should be able drop include/exclude and just remove that code. This was my reasoning.

Yes, this is right. I was more thinking of a two steps mechanism of inclusion-then-inclusion/exclusion per function based on the option declaration rather than relying on a filesystem distribution. But If you have that tested and working, I believe it's pretty good enough.

src/index.ts
```diff
@@ -216,13 +215,16 @@ export class EsbuildPlugin implements Plugin {
...
+  /** Link or copy extras such as node_modules or package.patterns definitions */
   async copyExtras() {
     const { service } = this.serverless;
 
-    // include any "extras" from the "include" section
-    if (service.package.include && service.package.include.length > 0) {
-      const files = await globby(service.package.include);
+    // include any "extras" from the "patterns" section
+    const globalPatterns = [
+      ...new Set([ ...(service.package.include || []), ...(service.package.patterns || []) ]),
```
olup, line 224: Just for code clarity, what do you think of putting this at the same place where you merge exclude and patterns? (l.180). Otherwise it's good for me.

kabo: Yeah, good point, I've updated this.

floydspace:
> I see that we use `glob` and `globby`. We might just stick to `globby` everywhere? that would save us a `stat` directory lookup in the packing phase (because globby have an option to output files only) and could speedup build as `globby` uses `fast-glob` which _sounds_ faster than glob

Works for me.

floydspace (commented):
@kabo thank you for the PR.
Please correct me if I'm wrong. to recap:
- now we handle serverless patterns along with include/exclude, so no breaking changes from our side (but I would leave mention of include/exclude options in docs)
- on individual packaging everything we include temporary goes to `__only_` folder and then packaged per func (not sure how the files appear in folders eventually if you not copy them, but just renaming variables)

kabo:
Given this setup in `serverless.yml`

```yaml
funcA:
  handler: src/funcA.handler
  package:
    patterns:
    - resources/some.txt
funcB:
  handler: src/funcB.handler
```

You'd end up with a build folder like this:

- __only_funcA/resources/some.txt
- funcA.js
- funcB.js

Then when it's zipped:

- funcA.zip containing `resources/some.txt` and `funcA.js`
- funcB.zip containing `funcB.js`

Hope that clarifies.

floydspace (approved):
Looks good to me. Thank you @kabo

olup (approved):

PR #112, opened 2021-04-11
feat(esbuild): support esbuild plugins

olup:
This is a naive implementation, but that might be all we need here.

I used the `plugins` key to let users pass on a path to the plugin definition file. Should it be named `pluginsDefinitionFile` or something in this taste? I think "plugins" is enough - but can be changed.


floydspace (approved):
Looks good. thank you @olup

Issue #117, opened 2021-04-20
Plugin breaks with serverless 2.36.0


taschmidt:
After upgrading serverless to the just-released version 2.36.0, started receiving the following error:
```
  TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
      at validateString (internal/validators.js:124:11)
      at Object.join (path.js:1039:7)
      at new EsbuildPlugin (<project dir>/node_modules/serverless-esbuild/dist/index.js:48:33)
```

The problem appears to be dependence on the `serverless.config.servicePath` prop here:
https://github.com/floydspace/serverless-esbuild/blob/753e4aede6e826adc77cf82847520f1f84f9bef8/src/index.ts#L75

This appears to have been renamed in 2.36.0:
https://github.com/serverless/serverless/releases/tag/v2.36.0

olup:
Thx a lot for reporting !

PR #118, opened 2021-04-20
fix(serverless): allow use of newer serviceDir serverless key

olup:
this should fix and close https://github.com/floydspace/serverless-esbuild/issues/117


floydspace (approved):
Looks good. Thank you @olup. Merging it

PR #119, opened 2021-04-20
feat(externals): add option to disable packing of external modules

mattjennings:
Adds an additional configuration option `packExternals` which will disable installing & bundling modules listed as external.

This is useful for if you want to exclude externals from esbuild but you don't need them installed & packed separately. My use-case was with serverless-offline as externals can be imported normally from node_modules, leading to a faster build. Another use-case could be if you're using lambda layers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,9 +66,17 @@ custom:
     packagePath: absolute/path/to/package.json # optional - by default it looks for a package.json in the working directory
 ```
 
+This behavior can be disabled via `packExternals`:
+
+```yml
+custom:
+  esbuild:
+    packExternals: false
+```
+
 ### Usign esbuild plugins
 
-*Note that the plugins API is still experimental : see [the documentation page](https://esbuild.github.io/plugins/)*
+_Note that the plugins API is still experimental : see [the documentation page](https://esbuild.github.io/plugins/)_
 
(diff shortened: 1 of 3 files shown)


floydspace:
Hi @mattjennings, thank you for your PR.
> The package specified in the `exclude` option is passed to `esbuild` as `external`, but it is not included in the function bundle either.

we already have the option `exclude`, will it cover your use case?

what can we do extra is making `exclude` option accept boolean values, so if it is `true` it excludes everything set as `external`. I have the reason for this: having set `exclude` and `packExternals` does not make sense together (exclude would have no effect), so it is why we can keep only one option, but I need second eyes from @olup on it.

mattjennings:
Ah yes, I hadn't considered the `exclude` option. I agree, finding a way to use `exclude` is a better idea. A boolean might be fine, otherwise maybe allowing pattern matching or `*` as a value to designate excluding everything.

floydspace:
> maybe allowing pattern matching or `*` as a value to designate excluding everything

also works for me 👍

vamche:
@floydspace @mattjennings I've opened https://github.com/floydspace/serverless-esbuild/pull/167 for pattern matching. Could you please give it a look?

Issue #120, opened 2021-04-21
1.10.3 breaks "sls invoke local"


taschmidt:
Upgraded to 1.10.3 to fix #117 which now works with serverless >= 2.36.0. However, it appears to have broken `invoke local`. I get a "cannot find module" error and it appears to be looking in `src/...` rather than `.esbuild/.build/src/...`:

```
Error: Cannot find module '/Users/schmidtt/w/video-tracking/src/list'
Require stack:
- /Users/schmidtt/w/video-tracking/node_modules/serverless/lib/plugins/aws/invokeLocal/index.js
- /Users/schmidtt/w/video-tracking/node_modules/serverless/lib/plugins/index.js
- /Users/schmidtt/w/video-tracking/node_modules/serverless/lib/classes/PluginManager.js
- /Users/schmidtt/w/video-tracking/node_modules/serverless/lib/Serverless.js
- /Users/schmidtt/w/video-tracking/node_modules/serverless/scripts/serverless.js
- /Users/schmidtt/w/video-tracking/node_modules/serverless/bin/serverless.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
    at Function.Module._load (internal/modules/cjs/loader.js:725:27)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at AwsInvokeLocal.invokeLocalNodeJs (/Users/schmidtt/w/video-tracking/node_modules/serverless/lib/plugins/aws/invokeLocal/index.js:743:33)
    at AwsInvokeLocal.invokeLocal (/Users/schmidtt/w/video-tracking/node_modules/serverless/lib/plugins/aws/invokeLocal/index.js:222:19)
    at Object.invoke:local:invoke [as hook] (/Users/schmidtt/w/video-tracking/node_modules/serverless/lib/plugins/aws/invokeLocal/index.js:49:47)
    at PluginManager.invoke (/Users/schmidtt/w/video-tracking/node_modules/serverless/lib/classes/PluginManager.js:551:20)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

taschmidt:
This could be an issue with core serverless >= 2.36.0. I encountered the same error in a different project that doesn't even use esbuild (webpack).

floydspace:
hi @taschmidt, thank you for the issue, we've merged the fix. If you have a chance could you please check if it works for you?

[...]

PR #122, opened 2021-05-02
fix(pack): servicePath not set for invoke local

aheissenberger:
I debugged the problem #120 and found that the function [invokeLocalNodeJS](https://github.com/serverless/serverless/blob/6d2ed840fb21d469d8e4a548a38ca9302d5a9d2c/lib/plugins/aws/invokeLocal/index.js#L745) will use `this.serverless.serviceDir` to set the path to the transpiled handler entry file.

I also compared this with [serverless-webpack](https://github.com/serverless-heaven/serverless-webpack/blob/a2de4d8a93c448a45e501f846f9232d774589578/lib/prepareLocalInvoke.js#L20) and they handle this similar.

@floydspace can you check if I missed to set this correctly when `service.package.individually` is set to `true`.

### DIFF
--- a/src/pre-local.ts
+++ b/src/pre-local.ts
@@ -1,6 +1,7 @@
 import { EsbuildPlugin } from '.';
 export function preLocal(this: EsbuildPlugin) {
   this.serviceDirPath = this.buildDirPath;
+  this.serverless.config.servicePath = this.buildDirPath;
   // Set service path as CWD to allow accessing bundled files correctly
   process.chdir(this.serviceDirPath);
 }


src/pre-local.ts
```diff
@@ -1,6 +1,7 @@
 import { EsbuildPlugin } from '.';
 export function preLocal(this: EsbuildPlugin) {
   this.serviceDirPath = this.buildDirPath;
+  this.serverless.config.servicePath = this.buildDirPath;
   // Set service path as CWD to allow accessing bundled files correctly
   process.chdir(this.serviceDirPath);
```
CorentinDoue, line 6: is this still required? invoke local from AWS serverless plugin seems to only use `this.serverless.config.servicePath`

aheissenberger: I am not sure - if you look at the code from the Webpack plugin there is still the `chdir`

floydspace:
Hey @aheissenberger thank you for contribution, I think this is the reason, I checked the serverless-offline plugin - it indeed relies on this option. merging it.

Issue #98, opened 2021-03-08
Memory usage grows exponentially of the esbuild process


thomaschaaf:
When using serverless-offline and making changes to files the memory usage of the esbuild process seems to grow exponentially.

I start out with about 1 GB of ram usage and after 10 changes the esbuild process is using 8 GB of ram.

olup:
I can confrim this. I am running into stack overflows continuously - and my colleagues on the project too.
It is probably linked to the alchemy between `chokidar` and `esbuild`.

olup:
From what I can see this does the leak: https://github.com/floydspace/serverless-esbuild/blob/ad3e93e5452122c17f110648c4f2e2a2bd0463c3/src/index.ts#L160 `true` option here is using incremental bundling. Maybe a lead.

olup:
incremental build does keep things in memory to speed up subsequent bundling: https://esbuild.github.io/api/#incremental. Could it be a leak in esbuild itself? Should we call `dispose` to free up resources from time to time ?

thomaschaaf:
@olup I think it's just used incorrectly. We need to call rebuild instead of initiating a new esbuild instance. That way rebuilding would be even faster aswell.

olup:
Yes true, otherwise the cache is a bit lost anyway. I'll flesh a pr tomorrow. Beyond this one subject, serverless offline, in watch mode, is terribly leaking so the rest of the process will end up eating the whole memory anyway. But we can improve what depends on us.

thomaschaaf:
I rewrote it to use rebuild. Sadly this does not resolve the memory issue. https://github.com/floydspace/serverless-esbuild/pull/101 so I think it might actually be an esbuild issue.

mattisdada:
I am reporting I've had similar issues, each build just linearly increases the amount of memory esbuild is using, if I was to hazard a guess, I'm thinking it has something to do with `node_modules` and this quote from the esbuild [docs](https://esbuild.github.io/api/#incremental):

> Files are stored in memory and are not re-read from the file system if the file metadata hasn't changed since the last build. This optimization only applies to file system paths. It does not apply to virtual modules created by plugins.

I have a feeling something is causing all the metadata change from ESBuilds perspective and it might not garbage collect old metadata (I imagine it doesn't watch for that). But I haven't looked at any of the code for `serverless-esbuild` so I might be completely wrong...

[...]

PR #123, opened 2021-05-02
Check for existing results to rebuild

anthony-zhou:
Currently, this plugin calls the ESBuild `build` method each time a file is changed.

As documented in #98, this behavior results in exponentially increasing memory usage as files are changed. Personally, I saw my memory usage (from the esbuild process) spike from an initial value of 600 MB to 2GB after just 5 changes. Ultimately, the memory usage results in a fatal error, where the JavaScript heap runs out of memory. 

## What I did

In this pull request, I've added a conditional check in the `bundle` method to call `rebuild` instead of `build` if there are existing build results.

I also set the `bundle` method to enable incremental builds by default. From what I understand, the way the ESBuild process works is as documented [here](https://esbuild.github.io/api/#incremental):

1. Create an initial build with `incremental: true`.
2. Rebuild as necessary by calling `rebuild` on the existing build result.

## Results

With the change in this PR, memory usage increases much more slowly. After about 10 file changes, the memory only increased to about 1GB (I noticed a linear increase of about 30MB each time). Further debugging is needed to determine the source of that slight memory usage increase, but it seems like this optimization addresses the majority of the existing increases in memory usage.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -229,9 +229,9 @@ export class EsbuildPlugin implements Plugin {
         const bundlePath = entry.substr(0, entry.lastIndexOf('.')) + '.js';
 
         if (this.buildResults) {
-          const { result: oldResult } = this.buildResults.find(({func: fn}) => fn.name === func.name);
-          await oldResult.rebuild();
-          return { result: oldResult, bundlePath, func };
+          const { result } = this.buildResults.find(({func: fn}) => fn.name === func.name);
+          await result.rebuild();
+          return { result, bundlePath, func };
         }
 
         const result = await build(config);
--- a/src/index.ts
+++ b/src/index.ts
@@ -201,7 +201,7 @@ export class EsbuildPlugin implements Plugin {
     }
   }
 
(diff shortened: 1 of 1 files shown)


floydspace:
Hi @anthony-zhou , appreciate your contribution and memory analyze. I see we have already similar PR opened https://github.com/floydspace/serverless-esbuild/pull/101, could you pls check it and make a new measurement with your test case? so we can kill 2 PRs if it successful.

anthony-zhou:
@floydspace After trying #101, the memory usage there is the same as before -- it increases exponentially (~3GB memory usage after 6 changes). I think the reason is that, in #101, the initial build is not created with the `incremental` flag set to true, meaning the `rebuild` method is never defined (as you mentioned in a comment on that PR). As such, the `rebuild` method never gets called.

As for the `esbuildCache` in #101, I don't know if that's necessary, since we're already storing the `buildResults` in `this.buildResults`.

floydspace:
@anthony-zhou okay merging it. Looks fine for me

Issue #121, opened 2021-04-25
How to create a "metafile"?


taschmidt:
I tried adding `metafile: meta.json` but got an error from esbuild that metafile must be a boolean. So I changed it to `metafile: true`. The build ran successfully but I was never able to find any output. Has anyone managed to analyze their esbuild bundle sizes?

floydspace:
hi @taschmidt, I checked the issue. As long as we are using esbuild's JavaScript API (not CLI), `metafile` option behaves a bit different, instead of defining a metafile name it receives a boolean value (according to `esbuild` source code, unfortunately it's not documented), and if the value is `true` the metadata is returned in the build result as JS object, so you can find your meta in the `result` variable here: https://github.com/floydspace/serverless-esbuild/blob/b5a9bdbe4be8b9decfda5590f92120d72ce8fe0e/src/index.ts#L229

taschmidt:
Yeah sorry, I did do some digging in esbuild and saw that it appears you can only do it via code. It might be handy to check here if metafile is true and, if so, dump it out to meta.json or something.

floydspace:
Hey @taschmidt , if it's still relevant for you, I completed the fix that will save metafile alongside the main bundle.

taschmidt:
Awesome! Thanks so much!

linonetwo:
Thanks, get it.

so we can have the solution:

```
const result = await esbuild.build({
  write: false,
  // we will have result.metafile later
  metafile: true,
  // other configs...
  entryPoints: packageJSON.tsFiles.map((tsFileName) => `./src/${tsFileName}.ts`),
  bundle: true,
  minify: false,
  sourcemap: process.env.CI ? false : 'inline',
  //....
});

// output bundle
for (let out of result.outputFiles) {
  await fs.mkdirp(path.dirname(out.path));
  await fs.writeFile(
    out.path,
    new TextDecoder()
      .decode(out.contents),
    'utf8',
  );
}

if (result.metafile) {
  // use https://bundle-buddy.com/esbuild to analyses
  await fs.writeFile('./dist/metafile.json', JSON.stringify(result.metafile));
}
```

PR #129, opened 2021-05-19
feat: write metafile when the metafile option is set to true

fargito:
Hello and thanks for this awesome plugin !

When debugging a performance issue on some of my lambdas, I found that it was not possible to output the meta files and therefore the debugging was a bit painful.

This is due to a specific implementation of the esbuild metafile option, that returns the meta in the result object of the build, but delegates the handling of it to its caller : https://esbuild.github.io/api/#metafile

Closes: https://github.com/floydspace/serverless-esbuild/issues/121

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -237,6 +237,10 @@ export class EsbuildPlugin implements Plugin {
 
         const result = await build(config);
 
+        if (config.metafile) {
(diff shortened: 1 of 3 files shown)


floydspace:
@fargito thank you for your contribution. Could you please also make sure it works with individual packaging, in this case, each lambda is packed in separate archive, and meta file will be generated and must be included per lambda as well
thank you.

fargito:
Hello @floydspace, It also works with individual packaging ! In both cases, a meta file is generated per lambda :)

PR #131, opened 2021-05-25
Fix bad example link in README.md

m14t:
Currently if you click the `See example folder for a minimal example.` link in the README.md it takes you to https://github.com/floydspace/serverless-esbuild/blob/master/example (singular), which is not a valid url.

Instead it should take you to https://github.com/floydspace/serverless-esbuild/blob/master/examples (plural)


floydspace:
hi @m14t , thanks for your PR, any size of contributions is appreciated.

PR #132, opened 2021-05-28
fix(sls file imports): Use generate config outside of the constructor

philiiiiiipp:
Currently the plugin accesses `this.serverless.service.custom?.esbuild` within the constructor. This does not work when imported variables are used, e.g. `${file(...`

```yml
esbuild:
  packager: npm
  bundle: true
  define: ${file(./serverless.js):define}
```

The `${file...} import is not resolved in the constructor but only when called after the constructor call.


floydspace (approved):
Hi @philiiiiiipp, looks good. thank you for the PR

Issue #136, opened 2021-06-10
Include node_modules when externals are populated via plugins (for eg., esbuild-node-externals)


vamche:
Kudos to the team which is behind this plugin and the packaging speeds it comes with! 🙇 

I have a question (feature-request?) for you. When the below code is packaged, the final deployment does not include the externals (node modules). I think it is because the externals are populated by a plugin in this case `esbuild-node-externals`.

https://github.com/vamche/sls-esbuild-node-externals/blob/4cc8843ed9c486727d3c3e89dabc1904ae7521ef/plugins.js#L3-L6
``` js
const { nodeExternalsPlugin } = require('esbuild-node-externals');


// default export should be an array of plugins
module.exports = [nodeExternalsPlugin()];
```

https://github.com/vamche/sls-esbuild-node-externals/blob/4cc8843ed9c486727d3c3e89dabc1904ae7521ef/serverless.yml#L15-L17
``` yml
  esbuild:
    plugins : ./plugins.js
    packager: npm
```

Sample repo: https://github.com/vamche/sls-esbuild-node-externals

Is there a way to include `node_modules` when externals are populated via plugins?

vamche:
FYI, `serverless-webpack` does this with `includeModules` config and `webpack-node-externals` plugin. 

I am happy to help/contribute to code if this feature can be introduced (if not already available). I would need some direction though. :)

floydspace:
Hi @vamche , thank you for the issue and the example. I cloned your repo and indeed the `esbuild-node-externals` plugin is excluding your dependencies from the bundle, but `serverless-esbuild` assumes that they are bundled.

I think it makes sense to implement a special case with the plugin, so if it is used we just get the list of modules determined by the plugin and put them to `external` to have packed.

If you really like to contribute, here is the direction: the plugin has exported function `findDependencies` you can reuse it to build a list of dependencies to be packed, and then the list must be used in our `packExternalModules` function which does the job. Should be an easy fix. Thank you 🙏

vamche:
Thank you @floydspace, I will give it a shot.

PR #138, opened 2021-06-11
feat(external): pack node_modules when esbuild-node-externals is configured

vamche:
Closes https://github.com/floydspace/serverless-esbuild/issues/136


vamche:
Closing as the file exists check is not working as expected.

PR #141, opened 2021-06-12
Feature/ignore non node functions

jcandrade:
Checking for non node functions and ignoring them, in case we have multiple runtimes per stack, like python.
Also, the plugin doesn't fail anymore if no node function was found, since we can manually deploy non node functions.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -127,19 +127,36 @@ export class EsbuildPlugin implements Plugin {
     };
   }
 
+  /**
+   * Checks if the runtime for the given function is nodejs.
+   * If the runtime is not set , checks the global runtime.
+   * @param {Serverless.FunctionDefinitionHandler} func the function to be checked
+   * @returns {boolean} true if the function/global runtime is nodejs; false, otherwise
+   */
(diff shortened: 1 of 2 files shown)


david-fontanals:
@floydspace, can you please merge and release this fix?

src/index.ts
```diff
@@ -127,19 +127,36 @@ export class EsbuildPlugin implements Plugin {
...
+   * Checks if the runtime for the given function is nodejs.
+   * If the runtime is not set , checks the global runtime.
+   * @param {Serverless.FunctionDefinitionHandler} func the function to be checked
+   * @returns {boolean} true if the function/global runtime is nodejs; false, otherwise
+   */
+  private isNodeFunction(func: Serverless.FunctionDefinitionHandler): boolean {
+    const runtime = func.runtime || this.serverless.service.provider.runtime;
+    const runtimeMatcher = providerRuntimeMatcher[this.serverless.service.provider.name];
+    const target = runtimeMatcher === null || runtimeMatcher === void 0 ? void 0 : runtimeMatcher[runtime];
+    return !!target;
```
samchungy, line 140:
```suggestion
    return Boolean(runtimeMatcher?.[runtime]);
```

PR #142, opened 2021-06-12
feat: keepOutputDirectory flag

taschmidt:
Copying over [this flag](https://github.com/serverless-heaven/serverless-webpack#keep-output-directory-after-packaging) from the serverless-webpack plugin. Can be very useful in diagnosing not only build issues but in tandem with the `metafile` option for analyzing bundle size.


taschmidt:
Bump

olup:
Agreed that this is indeed a helpful option to get. I will let @floydspace decide to merge this.

floydspace:
@taschmidt it's indeed could be helpful, at least not harmful at all. thanks for your contribution. merging it.

PR #147, opened 2021-06-17
fix(pack): function files with names whose prefix is same as other function names

vamche:
This is a fix.

When functions have names with prefix which is same as any other function name, their bundled js files are being excluded in the final artefact.

Ex: For an application having two functions with names `function-name` and `function-name-other`. The packaging (individually) is excluding the bundled `js` & `js.map` files for `function-name-other` in the final artefact.


vamche:
@floydspace Could you please review?

floydspace:
@vamche makes sense. thank you. merging it.

PR #148, opened 2021-06-17
feat(zip): utilize bestzip instead of archiver

### LINKED ISSUE #96: Use bestzip instead of archiver
https://www.npmjs.com/package/bestzip should make deployment quicker and use less memory. serverless-webpack has decided to use it too: https://github.com/serverless-heaven/serverless-webpack/pull/596/files

vamche:
Closes #96

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "typescript": "^4.0.3"
   },
   "dependencies": {
-    "archiver": "^5.2.0",
+    "bestzip": "^2.2.0",
     "chokidar": "^3.4.3",
     "esbuild": ">=0.8",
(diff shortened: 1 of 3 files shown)


olup (approved):
Tested on two of my projects, runs fine, thanks a lot ! I though our linting disallowed implicit any, but it's not the case. I think the pr can be merged as it is, but that would be best practice to add some simple typings for the bestzip function so that the import is not `any` typed.

floydspace (approved):

floydspace:
@vamche , @olup Appreciate guys, looks good to me as well, merging it.

PR #151, opened 2021-06-25
fix: resolve working dir issue with zipping code when not packaging i…

aaroncowie:
…ndividually

Hi, just having some issues locally with v1.14.0 where :
```
package:
  individually: false
```
causes the esbuild output to get written out successfully but then fails to zip with an error like
```
Serverless: Compiling to node12.22.1 bundle with esbuild...
Serverless: Compiling completed.
 
 Error ---------------------------------------------------

  Error: ENOENT: no such file or directory, stat 'c:\myrepopath\myreponame\src\functions\myfunctionname.js'

     For debugging logs, run again after setting the "SLS_DEBUG=*" environment variable.

  Get Support --------------------------------------------
     Docs:          docs.serverless.com
     Bugs:          github.com/serverless/serverless/issues
     Issues:        forum.serverless.com

  Your Environment Information ---------------------------
     Operating System:          win32
     Node Version:              15.5.0
     Framework Version:         2.48.0 (local)
     Plugin Version:            5.4.2
     SDK Version:               4.2.3
     Components Version:        3.12.0
```
This change seems to fix it for me locally. Feel free to reject if it doesn't work for other use cases!

### DIFF
--- a/src/pack.ts
+++ b/src/pack.ts
@@ -75,7 +75,7 @@ export async function pack(this: EsbuildPlugin) {
     )(files);
 
     const startZip = Date.now();
-    await zip(artifactPath, filesPathList);
+    await zip(artifactPath, filesPathList, this.buildDirPath);
     const { size } = fs.statSync(artifactPath);
 
     this.serverless.cli.log(


danmaas:
Hit the same issue here (and accidentally created my own PR before I noticed yours :) ). Fix works for me too.

vamche:
LGTM. Thank you for the fix @aaroncowie. 

I definitely overlooked this scenario in https://github.com/floydspace/serverless-esbuild/pull/148, sorry for that. 

@floydspace @olup Can we merge this one ASAP?

olup:
Works for me, I'll merge it.

PR #153, opened 2021-06-26
Fix zip() usage in non-individually packaging mode

### LINKED ISSUE #96: Use bestzip instead of archiver

danmaas:
This PR fixes a small bug introduced in PR #148 which fixed #96.

#148 updated one call site for `zip()` but the other call site, which handles the non-individually packaging mode, was not updated.

Without this patch, the `pack` step fails with the following error when your `serverless.yml` does not use the `package: individually` option:

```
zip error: Nothing to do! (try: zip --quiet --recurse-paths ${HOME}/myproject/.esbuild/.serverless/my-api.zip . -i src/publicApiHandler.js)
```

### DIFF
--- a/src/pack.ts
+++ b/src/pack.ts
@@ -75,7 +75,7 @@ export async function pack(this: EsbuildPlugin) {
     )(files);
 
     const startZip = Date.now();
-    await zip(artifactPath, filesPathList);
+    await zip(artifactPath, filesPathList, this.buildDirPath);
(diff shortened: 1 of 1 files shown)


danmaas:
Oops, sorry, this duplicates #151

PR #155, opened 2021-07-01
feat(packager): ability to run custom scripts after packager install

vamche:
This gives the ability to run custom scripts after packager install.

Use cases: remove/clean-up node modules etc.,

I have tested this with both `yarn/npm`, also w/ packaging individually `true/false`

You can find the reference script here: 
https://github.com/vamche/sls-esbuild-node-externals/blob/7a88127993acda4b12867245e8ac023cc68a6e13/serverless.yml#L22-L25

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,10 @@ custom:
   esbuild:
     packager: yarn # optional - npm or yarn, default is npm
     packagePath: absolute/path/to/package.json # optional - by default it looks for a package.json in the working directory
+    packagerOptions: # optional - packager related options, currently supports only 'scripts' for both npm and yarn
+      scripts: # scripts to be executed, can be a string or array of strings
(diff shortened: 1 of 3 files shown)


floydspace (approved):
thank you @vamche, looks good to me

vamche:
@olup Could you please review when you have some time?

olup:
Ok, I'll check tonight, but if @floydspace is good with it I have not much to add. On top of my head only, there was a PR recently about passing arguments to the install script, and I feel this one superspeeds the other one for the best.

vamche:
@floydspace What do you think? Can I merge this one?

floydspace:
> @floydspace What do you think? Can we merge this one?

Yes feel free to hit the button.

PR #165, opened 2021-07-22
Spelled out options specific to this plugin (not esbuild)

ffxsam:
I noticed `exclude` was mentioned briefly but kind of glossed over, then I realized it would be nice if the README just listed out the possible options for this plugin (not esbuild).


floydspace:
@ffxsam Hi Sam. it was initially a bad idea to put plugin specific options at the same level as esbuild options. In a future major releases, I would prefer having plugin options in root level and esbuild options under a separate key `esbuildOptions`. But anyway thank you for your valuable contribution.

PR #167, opened 2021-08-02
feat(exclude): pattern matching for exclude

vamche:


src/index.ts
```diff
@@ -228,7 +228,7 @@ export class EsbuildPlugin implements Plugin {
       this.rootFileNames.map(async ({ entry, func, functionAlias }) => {
         const config: Omit<BuildOptions, 'watch'> = {
           ...this.buildOptions,
-          external: [...this.buildOptions.external, ...this.buildOptions.exclude],
+          external: [...this.buildOptions.external, ...removeWildCardStrings(this.buildOptions.exclude)],
```
vamche, line 231: Since the plugin passes the packages included in `exclude` currently, I tried not to disturb that behavior. But the wild card/patterns don't make any sense to pass, so removing them here, including only the plain strings.

vamche:
@floydspace @olup Could you please review when you have some time?

vamche:
Hello @floydspace, @olup
Did you get a chance to review the PR?

floydspace:
Hi @vamche , I couldn't yet, was Covid-19 positive. Will check the PR this week. Thanks

vamche:
@floydspace Sorry to hear that, hope you are doing OK now. I wish you a speedy recovery 🚀

floydspace:
Hi @vamche , I tested your changes and got a couple of remarks. First I want to clarify how the `external` and `exclude` work together:
1. by default if `external` and `exclude` are not set, `esbuild` will bundle all references in output.
2. when you list libs in an `external` option, `esbuild` will not bundle them but instead will use node_modules packaging logic.
3. when we use `exclude` option, `esbuild` will do the same as in 2. + libs will not be packaged as node_moduels.

what I expect from this PR: the goal is to implement an ability to `exclude` libs from packaging as `node_modules` using wildcard pattern. so instead of listing all the libs in `exclude`, we would set `*` only (or any other pattern like `lod*sh` or whatever) and behavior from point 3 should persist. So we should search the libs according to pattern in package.json in packaging logic and do not copy them (exclude) to node_modules. 

But what I see now, it uses pattern to search libs in `external` list, so basically you have to set `externals` and `exclude` to have it somehow work.  But why to set externals if we `exclude` them anyway. Something wrong here. I would suggest keep it simple first, like implement `*` only, it would mean we `exclude` everything, so node_modules packaging logic is disabled, and if it's clear enough to make more complex logic with patterns, continue with it if it is really necessary.

Thank you.

vamche:
@floydspace Thanks for the review, it helped me to understand the plugin more. 

I've updated the PR, as you said, I kept it simple to support only '*' which should disable `node_modules` packaging. Could you please review it again?

Usage:
``` yaml
esbuild:
    exclude: "*"
```
or 

``` yaml
esbuild:
    exclude: 
      - "*"
```

vamche:
@floydspace Could you please review?

vamche:
Hello @floydspace, hope everything is OK on your side. I just wanted to check if you got a chance to review the PR after the changes?

vamche:
@floydspace Did you get a chance to look at the PR again?

vamche:
@floydspace could you please give this a look again?

floydspace (approved):
Hey @vamche , sorry that ignored it for a long time, I approve it. I will merge it as a minor fix

ericmaicon:
hey there, 

I am not sure if this is working properly. I setup exclude as the following:

```
exclude: '*'
```

check how is the size of my zip file when I run `yarn sls package`:

image

now I tried to remove all the things I have as dependencies on exclude:

```
exclude: [
        'pg',
        'reflect-metadata',
        'typeorm',
        'typeorm-naming-strategies',
        'unzipper',
        '@serverless/typescript',
        '@types/aws-lambda',
        '@types/faker',
        '@types/jest',
        '@types/node',
....
]
```

You can see that I needed to list even my dev dependencies here. After that, this is what I got:

image

The goal wans't to list all of it?!

The versions I am using is the following:
```
"serverless": "^2.23.0",
    "serverless-esbuild": "^1.23.1",```

Thanks in advance.

**Comment by AlisonVilela** [comment]
[ericmaicon](https://github.com/ericmaicon) Same here!

wildcard doesn't work, I need to put all the libraries I'm using one by one

**Comment by samchungy** [comment]
I'll take a peek over the weekend 👀

**Comment by vamche** [comment]
@ericmaicon @AlisonVilela This implementation requires the node modules to be set as externals as well.

Here is a sample configuration.
`serverless.yml`
``` 
  esbuild:
    plugins: ./esbuild-plugins.js
    exclude: '*'
```

`esbuild-plugins.js`
```
const { nodeExternalsPlugin } = require('esbuild-node-externals');

module.exports = [
  nodeExternalsPlugin(),
];
```

Please let me know if you have questions. thank you!

PR #171, opened 2021-08-31
fix(package): fix wrong argument to get package's root

olup:
A problem arouse in one of my project : mongoose was required by a dependency with a specific version. We put mongoose as external because it relies on system code.
But as mongoose was not listed as dependency of our project, its version was not caught by the plugin ; as such a later, incompatible version of mongoose was packaged.
This PR changes the way we get the dep versions by checking the module package.json found either in node_modules or the root node_modules (in case of a workspaced project).
It also changes the arguments to get the project root (a package.json is always found even if it is not the root).

### DIFF
--- a/src/pack-externals.ts
+++ b/src/pack-externals.ts
@@ -85,35 +85,62 @@ function getProdModules(
 
   // Get versions of all transient modules
   forEach(externalModule => {
-    const moduleVersion = packageJson.dependencies[externalModule.external];
+    // (1) If present in Dev Dependencies
+    if (
+      !packageJson.dependencies[externalModule.external] &&
+      packageJson.devDependencies[externalModule.external]
+    ) {
+      // To minimize the chance of breaking setups we whitelist packages available on AWS here. These are due to the previously missing check
+      // most likely set in devDependencies and should not lead to an error now.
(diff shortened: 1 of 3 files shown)


src/pack-externals.ts
```diff
@@ -85,35 +85,62 @@ function getProdModules(
 
   // Get versions of all transient modules
   forEach(externalModule => {
```
olup, line 87: @floydspace quite bit of refacto here - can you have a quick look ?

src/utils.ts
```diff
@@ -64,35 +64,36 @@ export function spawnProcess(
...
 
-  return findUp(name, path.dirname(absoluteDirectory));
+  return findUp(names, path.dirname(absoluteDirectory));
 }
 
 /**
  * Forwards `rootDir` or finds project root folder.
  */
 export function findProjectRoot(rootDir?: string): string | undefined {
-  return rootDir ?? findUp(['yarn.lock', 'package-lock.json', 'package.json']);
```
olup, line 95: There is a `package.json` file even in workspace - so this had to be removed

floydspace (approved):
Looks okay to me, thank you

PR #183, opened 2021-09-14
Always include files that are in patterns

BenJanecke:
Before this change if a directory or file was in node_modules for
example the `.prisma` directory that gets created by prisma but not marked
as a dep or peer dep it would be removed.

Adding this directory to patterns did not include it in the artifact.

This change seeks to always include files that are present in patterns.

### DIFF
--- a/src/pack.ts
+++ b/src/pack.ts
@@ -104,6 +104,10 @@ export async function pack(this: EsbuildServerlessPlugin) {
     ? await packager.getProdDependencies(this.buildDirPath)
     : {};
 
+  const packageFiles = await globby(
+    this.serverless.service.package.patterns
+  );
+
   // package each function
   await Promise.all(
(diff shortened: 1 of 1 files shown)


floydspace:
Hi @BenJanecke thank you for the PR, I'd like to test it, so it would take time to pick it.
I'm wondering if you would generate `.prisma` outside `node_modules`, would it include it?

BenJanecke:
Hi @floydspace yes this is possible by passing a generated path.

However I think this change will be useful to people that want to include other modules in `node_modules` that might not be set up correctly i.e. not marked correctly as peer-dependencies. 

The proper fix of course would be adding the packages as strong dependencies but this would be very useful as a stop gap.

Also there are other libraries that put their generated code in `node_modules` that could also benefit from this

floydspace:
@BenJanecke yes, it makes sense. I will test it as soon as I'm able to focus on it. thank you

Issue #186, opened 2021-09-17
Generated zip files always have new timestamps and force stack updates


troyready:
Every invocation of [bestzip](https://github.com/nfriedly/node-bestzip) generates a zip file with a different sha256 hash because of the contents' timestamps. Similar issue [exists for serverless-webpack as well](https://github.com/serverless-heaven/serverless-webpack/issues/881).

When Serverless is creating zips, it avoids this [by explicitly setting the files' timestamps to the epoch](https://github.com/serverless/serverless/blob/v2.59.0/lib/plugins/package/lib/zipService.js#L101). Bestzip [could copy this](https://github.com/nfriedly/node-bestzip/blob/v2.2.0/lib/bestzip.js#L122-L126) for platforms without a `zip` executable, but it would require a different approach for when it shells out to `zip` (which doesn't have a flag to override timestamps).

troyready:
Some solutions that come to mind:
* Switch to a deterministic zip provider
  * serverless-webpack is likely going back to `archiver`
* Modify the files' timestamps before zipping them
  * Something like the javascript equivalent of `find . -exec touch -t 197001010000 {} +`

I'm partial to switching to `archiver`, but curious for your thoughts/preference.

olup:
Very interesting. I'd second going back to archiver, as I am no fan of bestzip. 

However, as this was an improvement for some heavy duty use cases, it might be a good idea to follow the (still unmerged it seems) PR of serverless webpack, and add a `nativeZip` option to let user opt-in best zip.

floydspace:
Hey, @olup, I can't remember exactly what benefit we got from migrating to bestzip? are those improvements worth the issues we have currently? I'd also vote for reverting back to `archiver`.
This week I went through the [zip-it-and-ship-it](https://github.com/netlify/zip-it-and-ship-it) by Netlify, they also are using archiver.

olup:
The original issue should be here somewhere. Someone argued, with benchmarks, that it made a huge performance difference when zipping many many functions. My concerns with bestzip were low maintenance and no typings.

Issue #150, opened 2021-06-23
Regression of #110, package.patterns are ignored


kabo:
It seems #110 is back.

serverless.yml:

```yaml
functions:
  myFn:
    package:
      patterns:
        - resources/some.txt
```
 
Running `serverless package` with serverless-esbuild v1.10.6 works as expected.

```
$ cd .serverless
$ unzip myFn.zip
Archive:  myFnl.zip
  inflating: resources/some.txt
  inflating: myFn.js
```

But with v1.10.7 and onwards it's no longer working, the extra file is missing in the zip.

```
$ cd .serverless
$ unzip myFn.zip
Archive:  myFn.zip
  inflating: myFn.js
```

kabo:
Ah, wait, it's not that simple. I took a look at the commit between 1.10.6 and 1.10.7. The issue is here: https://github.com/floydspace/serverless-esbuild/commit/e504e48220ea910b5e8fd55aef58166e8b14821b#diff-c0b27c44c8eee1df38854f691661101b9de7cfcf5b832817bd6012983d3aa9beR110

In my serverless.yml:

```yaml
provider:
  stage: test
```

I then run `yarn serverless package -s prod`. In 1.10.6 the zip had "prod" in the filename, from 1.10.7 it has "test" in the filename. If I change provider.stage to prod in my serverless.yml it works again. So it's failing to take the commandline option into account, that's the real bug, that the file fails to be included is just a downstream bug from that.

kabo:
No, wait, the above comment only makes version 1.10.7 - 1.13.1 work. Version 1.14.0 doesn't include the extra file.

olup:
I can reproduce this if like you I am adding the pattern option on an individual function. However when I set the pattern on the global packaging options it works (and include the file on every function). Could you confirm that on your end ?

kabo:
Yeah, I can confirm that works. But it's pretty useless when you have one lambda that requires one big file, and you want to keep all the other lambdas in the service that don't need that file as small as possible.

olup:
Yeah, agreed. But now I can look for a fix ;-)

olup:
I am working on the bug. Making a note here:

The regression seems to come from #148 - the move to Bestzip. Bestzip works in a different way than what we had previously - most importantly, it does not allow to zip a file from a path to another path inside the zip. the individual patterns or file include relied on the creating of a temporary folder holding the desired files so that each function only get whats it was asking for. 

[...]

PR #187, opened 2021-09-18
feat(zip): make bestzip an off-by-default option

troyready:
bestzip preserves file timestamps, which causes function zips to be different after every esbuild.

Because there are scenarios where native js zipping is unacceptably slower than native system zip, the option to use it is preserved as an option.

Fixes: #186

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,18 @@ module.exports = (serverless) => {
 };
 ```
 
+### Native Zip
+
+If you wish to use your system's `zip` executable to create archives (can be significantly faster when working with many large archives), set the `nativeZip` option:
+
+```yml
+custom:
+  esbuild:
+    nativeZip: true
+```
+
+**NOTE:*** This will produce non-deterministic archives which causes a Serverless deployment update on every deploy.
+
 ## Usage
 
 ### Automatic compilation
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   "devDependencies": {
     "@commitlint/cli": "^9.1.1",
     "@commitlint/config-conventional": "^9.1.1",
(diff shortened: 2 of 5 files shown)


troyready:
FYI @vamche -- I'm proposing a change here that would require a small config change to preserve your use of system `zip`.

vamche:
Hey @troyready, thanks for the PR. Looks good to me. 

I have suggested a similar approach in the past but we didn't take it forward because of some concerns. Here is the related thread/comment https://github.com/floydspace/serverless-esbuild/issues/150#issuecomment-872010032. Could you please check with @olup?

vamche (approved):

troyready:
Thanks for the background, that is helpful.

FWIW, in regards to that exact comment: we wouldn't be able to use `bestzip.nodeZip` as is, because it suffers from the same timestamp issue. It might not be unreasonable to check in with them on supporting it.

To [olup's comments](https://github.com/floydspace/serverless-esbuild/issues/150#issuecomment-872573660), I'm fine either way. I do like the simplicity of just having a single implementation (without another config option), but it's going to be a matter of opinion on whether that's the right tradeoff.

floydspace (approved):
@troyready Looks good to me as well. thanks that documented it. as soon @olup approves as it concerns his comment, I merge it

olup:
Well, I am partial to a simple implementation. And for me, the main thing is to have reliable timestamps - a problem that I did not know about before. However because the native zip seems to be an important feature for some users, and as we started supporting it, I am fine with an optional parameter that lets one opt-in native zipping (reliable timestamps should stay the default).

So I am good with this PR.

PR #197, opened 2021-09-23
feat: ignore external module if not declared in package.json

### LINKED ISSUE #196: external behaviour
## Description

The s3 client recently bundled an optional peer dep package called @aws-sdk/signature-v4-crt and I get the following error:
image

Going by the comments, I added the package to my `external` list. I do not have the plugin listed in my dependencies for package.json, however, the plugin still tries to installs it.

```
external
  - @aws-sdk/signature-v4-crt
 ```
image
image

I'm not sure if I'm misunderstanding the documentation but going by this bit here

```External Dependencies
Packages that are marked as external and exist in the package.json's dependencies will be installed and included with your build under node_modules. You can configure how these are installed:
```

is it right for me to believe that if it is not in my package.json, esbuild should not install it at all? In my opinion that sounds like a reasonable thing to assume and an effective way of not bundling an optional peer dep. The esbuild documentation even states that as the purpose of external. If this is the case -> I'm happy to dig into the code and implement it.

```
This has several uses. First of all, it can be used to trim unnecessary code from your bundle for a code path that you know will never be executed.
```

If that is not the case -> is there a way I can achieve this as I know the s3Client will not hit that section of code and I would like to ignore installing and packaging this bit of code.

samchungy:
## Why we need this change

Resolves #196

If you have a package listed in `external` which is *not* in the package.json dependencies, the plugin will try and install it anyway.

The documentation states the following:

```
Packages that are marked as external and exist in the package.json's dependencies will be installed and included with your build under node_modules.
```

## Change
- aligns the behaviour of `external` with the documentation.

### DIFF
--- a/src/pack-externals.ts
+++ b/src/pack-externals.ts
@@ -85,7 +85,20 @@ function getProdModules(
 
   // Get versions of all transient modules
   forEach((externalModule) => {
-    // (1) If present in Dev Dependencies
+    // (1) If not present in Dev Dependencies or Dependencies
+    if (
+      !packageJson.dependencies[externalModule.external] &&
+      !packageJson.devDependencies[externalModule.external]
+    ){
+      this.options.verbose &&
+        this.serverless.cli.log(
+          `INFO: Runtime dependency '${externalModule.external}' not found in dependencies or devDependencies. It has been excluded automatically.`
+        );
+        
+      return;
+    }
+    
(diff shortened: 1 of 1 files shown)


floydspace (approved):
Hi @samchungy .  Looks good. Thank you for the improvement. Merging it

PR #198, opened 2021-09-24
fix(pack): local lockfile resolved at wrong location

olup:
I had a bug today where yarn would try to install my externals with --frozen-lockfile, even from a service in a monorepo. Checked the code, and actually, we were resolving the lockfile from the root package json path - whereas the original PR for this feature stipulates that the lockfile should be sought at the local level - thus determining if we should use frozen lockfile or not.

This will not affect users out of a monorepo.

I am fixing this now, But I really think we should take some time to review the all packaging step, as it has become quite complex PR after PR. But time is hard to find :-).

### DIFF
--- a/src/pack-externals.ts
+++ b/src/pack-externals.ts
@@ -205,14 +205,14 @@ export async function packExternalModules(this: EsbuildServerlessPlugin) {
       packagePaths: findPackagePaths(),
       allowList: [],
     });
-  }  
+  }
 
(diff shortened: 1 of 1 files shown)


olup:
Hey @floydspace my team kind of really need this to work. Should I merge this? @vamche any oppositions?

floydspace (approved):
Go ahead

PR #201, opened 2021-09-29
feat: set up concurrency on building the lambdas

### LINKED ISSUE #200: Limit concurrency on esbuild packaging
Hi,

For some use cases, where there is limited system memory available (i.e. GitHub Actions 7GB), esbuild can run out of memory and crash when building a big serverless project (i.e. 50+ lambdas). For those cases, and according to the esbuild developer, it is recommended to do batching:

- https://github.com/evanw/esbuild/issues/320

In order to do so, and if I am not mistaken, the serverless-esbuild pack on here:

- https://github.com/floydspace/serverless-esbuild/blob/master/src/pack.ts#L113

Could use a library like [p-limit](https://github.com/sindresorhus/p-limit) or [p-queue](https://github.com/sindresorhus/p-queue) to limit the concurrent builds of lambdas, and configure the limit through an option in serverless.yml, for example under the key _concurrency_

Is that something that would make sense for this library or is it out of the scope of it?

If it would indeed make sense, is it something you would consider doing or would you accept a PR?

JoaquinFernandez:
## What has been implemented?

For some use cases, where there is limited system memory available (i.e. GitHub Actions 7GB), esbuild can run out of memory and crash when building a big serverless project (i.e. 50+ lambdas).

In order to overcome that I've added the esbuild option _concurrency_ to only allow as much as _concurrency_ promises running at the same time when building the lambdas.

I could help with the tests, and indeed I tried to work on them, but there is a lot to reestructure in the index.ts before unit tests start to make sense.

Closes #200

## Steps to verify

- Run _serverless package_ command with the esbuild option _concurrency_ as an integer
  - It should limit the concurrent building of the service lambda's to that amount

- Run _serverless package_ command with the esbuild option _concurrency_ as a string
  - It should throw an error with the message:
    - TypeError: Expected `concurrency` to be an integer from 1 and up or `Infinity`, got `a` (string)

- Run _serverless package_ command with **no** esbuild option _concurrency_
  - It should concurrently build all of the service lambda's

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,6 +2,7 @@ import { build, BuildResult, BuildOptions, Plugin } from 'esbuild';
 import * as fs from 'fs-extra';
 import * as globby from 'globby';
 import * as path from 'path';
+import * as pMap from 'p-map';
 import { concat, always, memoizeWith, mergeRight } from 'ramda';
 import * as Serverless from 'serverless';
 import * as ServerlessPlugin from 'serverless/classes/Plugin';
@@ -35,6 +36,7 @@ export interface PackagerOptions {
 }
 
 export interface Configuration extends Omit<BuildOptions, 'nativeZip' | 'watch' | 'plugins'> {
+  concurrency: number;
   packager: 'npm' | 'yarn';
   packagePath: string;
   exclude: '*' | string[];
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -244,54 +246,65 @@ export class EsbuildServerlessPlugin implements ServerlessPlugin {
...
+      this.buildResults = await pMap(this.rootFileNames, bundlePromise, {
+        concurrency: this.buildOptions.concurrency,
+      });
+    } else {
+      this.buildResults = await Promise.all(
+        this.rootFileNames.map(async (bundleInfo) => bundlePromise(bundleInfo))
+      );
+    }
+    this.serverless.cli.log('Compiling completed.');
+    return this.buildResults.map((r) => r.result);
```
JoaquinFernandez, line 307: These are mostly whitespace changes, (since I had to indent them correctly with the changes), please look at the PR hiding the whitespace changes

PR #224, opened 2021-11-03
Add docs for how to mark individual packages as external

gavynriebau:
In some situations it can be useful to include only some dependencies as external but allowing other dependencies to be bundled, one such example is when creating AWS Edge lambdas which have limit on deployment package sizes.

This commit adds to the docs an example of how to do this which isn't otherwise immediately obviously without going to look at example serverless.yml files.


floydspace (approved):
@gavynriebau thank you, any improvement is appreciated

PR #229, opened 2021-11-11
fix: filtering in pack.ts

### LINKED ISSUE #195: Serverless Offline path does not work with function level package pattern
I am trying to run the following in a function in Serverless Offline

```
  const buffer = await readFile(
    __dirname + '/user-data.sh'
  );
```
but I will receive a file not found error.

```
package:
  individually: true

functions:
  start-server:
    package:
      patterns:
        - src/functions/start-server/luser-data.sh
 ```

But the file will be not found because it looks like serverless-esbuild will output the file to __only-serverless-dev-start-server/functions/start-server/user-data.sh folder instead of src/functions/start-server/user-data.sh

It looks like the package version works fine because it copies it over but it does not work with serverless-offline.

My work around is setting the top level package to include the file instead.

capaj:
fixes #195

also tested on real project where we have patterns on individual lambda functions and it works fine

### DIFF
--- a/src/pack.ts
+++ b/src/pack.ts
@@ -17,7 +17,7 @@ import * as semver from 'semver';
 import { EsbuildServerlessPlugin, SERVERLESS_FOLDER } from '.';
 import { doSharePath, flatDep, getDepsFromBundle } from './helper';
 import * as Packagers from './packagers';
-import { IFiles } from './types';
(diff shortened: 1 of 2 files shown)


floydspace (approved):
Hey @capaj . thank you for the fix. Looks good to me.
Happy to see a unit test here 👍

PR #231, opened 2021-11-16
feat: added disableIncremental option to completely turn off incremental compilation

gurushida:
**Motivation: preventing a memory leak**

When used in combination with serverless-offline on a typescript codebase of about 200.000 lines of code, my colleagues and I have observed a memory leak that caused in our case about 500Mb to be consumed on each code change triggering a recompilation (as described in  #98). After some experiments with serverless-esbuild's code, it turned out that totally disabling incremental compilation made the problem disappear. Since esbuild is extremely fast, giving up incremental compilation did not affect our development speed, which is why we want to propose to add an option to do that.


floydspace (approved):
Hi @gurushida , thanks for your contribution, I think it's good to have it let me merge.

PR #236, opened 2021-11-30
fix: allow relative path for function handler

### LINKED ISSUE #234: Super super super slow to create Zip files
I have 20+ small functions.

```JavaScript
esbuild: {
  packager: 'npm',
  bundle: true,
  minify: false,
  sourcemap: false,
  treeShaking: true,
  nativeZip: true, // 👈 this helps a little but still very slow - in fact without it it mostly fails but with no meaningful error
  concurrency: 10, // 👈 makes no difference
  external: ['migrate-mongo', 'pdfkit', 'pdf-lib', 'jsonwebtoken'],
}
```

Running on a Mac:

image

Any ideas?

I've also noticed that the generated Zip files are all the same size - this seems weird but it looks like it's putting everything in the Zip. Am I missing something here?

### LINKED ISSUE #235: Package Individually breaks if lambda paths are prefixed with ./
Absolutely LOVE this plugin. Truly amazing, and a massive improvement over webpack.

Unfortunately the package individually option has stopped working. Basically this issue: https://github.com/floydspace/serverless-esbuild/issues/79 

Except it's broken again.

---

For anyone coming here with the same problem: your best workaround will be to just deploy each function individually in a custom npm script.  Ie

```
sls deploy function -f func1 && sls deploy function -f func2 && sls deploy function -f andSoOn
```

---

EDIT

This only happens when you have paths prefixed with `./`, ie

```
handler: ./handlers/scanBook.handler
```

whereas

```
handler: handlers/scanBook.handler
```

works perfectly.

samchungy:
Resolves #235
Resolves #234

This allows for users to set the function handler using the relative path operator: eg.

```
func1:
  handler: ./file1.handler
```

### DIFF
--- a/src/helper.ts
+++ b/src/helper.ts
@@ -9,7 +9,7 @@ export function extractFileNames(
   cwd: string,
   provider: string,
   functions?: Record<string, Serverless.FunctionDefinitionHandler>
-): { entry: string; func: any, functionAlias?: string }[] {
+): { entry: string; func: any; functionAlias?: string }[] {
   // The Google provider will use the entrypoint not from the definition of the
   // handler function, but instead from the package.json:main field, or via a
   // index.js file. This check reads the current package.json in the same way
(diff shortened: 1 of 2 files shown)


floydspace (approved):
Hi @samchungy thank you for covering it with tests.
The fix looks good. Will merge it immediatelly.

PR #238, opened 2021-12-04
ci: update workflows

samchungy:
This repo might get more contributors in the future, add a workflow for PRs so that test or lint issues do not need to be resolved when the release workflow runs.

- Adds a PR workflow. (adds lint warning annotations, so I addressed these below)
- Adds a package lock file
- Changed `any` type to `unknown` or a specific type where I could.

(Prettier formatted some of the files)


samchungy:
Should we add a lockfile?

src/pack-externals.ts
```diff
@@ -124,69 +124,69 @@ function getProdModules(
...
+        this.options.verbose &&
+          this.serverless.cli.log(
+            `Adding explicit non-optionals peers for dependency ${externalModule.external}`
           );
-          Array.prototype.push.apply(prodModules, peerModules);
-        }
-      } catch (e) {
-        this.serverless.cli.log(
-          `WARNING: Could not check for peer dependencies of ${externalModule.external}`
+        const peerModules = getProdModules.call(
```
samchungy, line 174: Nothing here has actually changed, just prettier changed the indentation

floydspace:
> Should we add a lockfile?

Hey @samchungy 
I agree we need to lock the versions, due to the previous deployment issue.

The initial idea was to always get the latest `esbuild` when installed by user, but it's not safe in terms of breaking changes.

I think it's even better to use `esbuild` as a peer dependency. but we have to force users install it together with this plugin:
`npm install serverless-esbuild esbuild -D`, any thoughts?

floydspace (approved):

samchungy:
> > Should we add a lockfile?
> 
> Hey @samchungy I agree we need to lock the versions, due to the previous deployment issue.
> 
> The initial idea was to always get the latest `esbuild` when installed by user, but it's not safe in terms of breaking changes.
> 
> I think it's even better to use `esbuild` as a peer dependency. but we have to force users install it together with this plugin: `npm install serverless-esbuild esbuild -D`, any thoughts?

I think this makes sense, Question is what version we want to restrict it to? Seeing as esbuild are pretty good with versioning and `0.12.0` has been out for a good 5-6 months I'm thinking `^0.12.0`.

floydspace (approved):
@samchungy approving it, thanks for contribution.
Regarding peer dependency.
I think we should use >=0.8 < 0.15, as it's time proved that >=0.8 works, and additionally restrict to install version higher the current one 0.14, to avoid possible breaking changes in the next version. and time to time shift the version to the right)

PR #240, opened 2021-12-06
feat: set esbuild as a peer depedency

samchungy:
`esbuild` is now a peer dependency. Please install `esbuild` as a dev dependency for your projects.


samchungy:
From our convo in https://github.com/floydspace/serverless-esbuild/pull/238

floydspace:
@samchungy oh, I guess this commit will update the package to v2, which I wouldn't feel like doing.
It should be really something major to update to the v2, IMO.

floydspace:
it'd be nice to fix readme as well.

samchungy:
> @samchungy oh, I guess this commit will update the package to v2, which I wouldn't feel like doing. It should be really something major to update to the v2, IMO.

yeah fair enough. I guess we can just mention it.

floydspace (approved):
@samchungy  thank you, merging it

Issue #128, opened 2021-05-13
Name of lambda function causing issue when esbuild is packaging files


danielhorne:
Hi Team

Firstly this is an awesome plugin.

This is an issue/feature request

When I try package my serverless project I get the following issue  _Error: ENOENT: no such file or directory, open 'C:\Users\****\repos\databus-serverless\.esbuild\.serverless\r-#{AWS::AccountId}-npr-ds-databus-hello.zip'_

The issue comes about because I am trying to add a prefix to my lambda name using the serverless-pseudo-parameters plugin. 
I have pasted my serverless config below
If I remove the _name_  field from the lambda config then your library defaults to the object property in the serverless config.

```
{
    service: "databus-serverless",
    frameworkVersion: "2.35",
    provider: {
        name: "aws",
        tags: globalTags,
        profile: "developerrole",
        region: "eu-west-1",
        runtime: "nodejs14.x",
        lambdaHashingVersion: 20201221,
        deploymentBucket: {
            name: "${self:custom.prefix}-global-auto-scaling-s3",
        },
        environment: {
            NODE_PATH: "./:/opt/node_modules",
        },
    },
    custom: {
        accountId: "#{AWS::AccountId}",
        account: "${opt:account, self:custom.accountId}",
		esbuild:{
			bundle:true,
			minify:false,
			sourcemap:true
		},
        region: "${opt:region, self:provider.region}",
        stage: '${opt:stage, "npr"}',
        prefix: "r-${self:custom.account}-${self:custom.stage}-ds-databus",
    },
    package: {
        excludeDevDependencies: true,
        individually: true
    },
    plugins: [
	"serverless-esbuild",
        "serverless-pseudo-parameters",
        "serverless-deployment-bucket",
        "serverless-offline",
    ],
    functions: {
        Hello: {
			name: "${self:custom.prefix}-hello",
            handler: "functions/hello/index.handler",
            role: "arn:aws:iam::${self:custom.account}:role/${self:custom.prefix}-iam-lambda-bifrost-role",
            layers: [
                "arn:aws:lambda:${self:custom.region}:${self:custom.account}:layer:modules:1",
            ]
        }    
  }
}
```

Is there or could there be a flag to enable this naming functionality. For example by default the build folders are named using the object property rather than the name field in function definition. In order to use the lambda name for the build folders there would be an options as follows:
```
esbuild:{
  useLambdaName:true
}
```

floydspace:
[...]

PR #249, opened 2021-12-21
fix: correct package individually patterns functionality

### LINKED ISSUE #246: `patterns` not working as expected for me - files missing from `.zip`
Hi.

Using version `1.23.1`.

I'm trying to get some files included in my deployment bundle - I have these files in `migrations-core` alongside my serverless function handlers.

image

On these functions only, I have this:

```JavaScript
package: {
  patterns: ['**/migrations-core/*'],
},
```

When the code is being bundled, I can see these `__only` folders with the files and it refers to the the 2 expected functions.

image

However, these files are never included in the `.zip` file.

Am I missing something here - how can I get these files included? This is my config.

```JavaScript
  esbuild: {
    bundle: true,
    minify: true,
    treeShaking: true,
    sourcemap: false,
    nativeZip: true,
    concurrency: 20,
    external: ['mongoose', 'winston', 'migrate-mongo', 'pdfkit', 'pdf-lib'],
    format: 'cjs',
},
```

samchungy:
Fixes #246

https://github.com/floydspace/serverless-esbuild/blob/fb8345d3f596fad4eb0452780e1cb4572cc09b9f/src/pack.ts#L163-L165

The issue here is that stage can be specified with `this.options.stage` which does not overide the `this.serverless.service.provider.stage ` plus the `copyExtra` function creates the output folder based on the function's name.

https://github.com/floydspace/serverless-esbuild/blob/fb8345d3f596fad4eb0452780e1cb4572cc09b9f/src/index.ts#L375-L377

## Change
The output of package individually zip files will now use the functionAlias as the zip name. eg.

```
function1:
  name: helloFunction
  handler: function.handler
```

will produce `function1.zip`. This matches how the Serverless Framework names their zips.


jest.config.js
```diff
@@ -1,7 +1,8 @@
 const config = {
     preset: 'ts-jest',
     verbose: true,
-    transform: {}
+    transform: {},
+    testPathIgnorePatterns: ['dist'],
```
samchungy, line 5: running `npm test` after `npm run build` would make Jest try to run the `.test.js` in the dist folder.

src/index.ts
```diff
@@ -15,10 +15,7 @@ import { pack } from './pack';
 import { preOffline } from './pre-offline';
 import { preLocal } from './pre-local';
 import { trimExtension } from './utils';
-
-export const SERVERLESS_FOLDER = '.serverless';
```
samchungy, line 19: https://github.com/floydspace/serverless-esbuild/blob/fb8345d3f596fad4eb0452780e1cb4572cc09b9f/src/index.ts#L441

Because of this line for some reason when the `pack.test.ts` file ran these variables were coming as `undefined`. Moved to their own file.

samchungy: Alternatively we could declare them on the class itself. I'm open to suggestions.

chearmstrong:
Nice one -  makes sense! Thanks @samchungy.

src/index.ts
floydspace: it would be good to have `constants.ts` to keep all the static values there. 
`folders` is too narrow, imo

src/pack.ts
```diff
@@ -159,11 +159,8 @@ export async function pack(this: EsbuildServerlessPlugin) {
 
   // package each function
   await Promise.all(
-    buildResults.map(async ({ func, functionAlias, bundlePath }) => {
-      const name = `${this.serverless.service.getServiceName()}-${
-        this.serverless.service.provider.stage
-      }-${functionAlias}`;
-
+    buildResults.map(async ({ func, bundlePath }) => {
+      const fnName = func.name;
```
floydspace, line 163: @samchungy 
Please consider this PR, in `v1.10.7` I made the opposite change due to some issue with custom lambda naming, maybe you could workaround this so both cases would work?

floydspace (changes requested):
Please verify that this issue #128 will not bubble up again

src/pack.ts
samchungy: Nice good thing I requested your review 😊. I think it'll just be adjusting `copyExtra` or this one to use the same syntax.

samchungy: Okay @floydspace  so I looked at the above linked issue and it looks like they were using https://www.npmjs.com/package/serverless-pseudo-parameters which is now deprecated as Serverless Framework 1.5.0 does this behaviour natively. It looks like Serverless also parses this before we package which is nice which means we shouldn't have the same issues as before.

However, of course if we continue with this serverless-pseudo-parameters will not work, but in my opinion that sounds okay to me as it is deprecated. I tested this with `${aws:region}`, `${aws:accountId}` as well as a `${self:custom.name}` and it appears to work fine.

image

Option 2:

If we would like to keep that behaviour I could always change this line

https://github.com/floydspace/serverless-esbuild/blob/fb8345d3f596fad4eb0452780e1cb4572cc09b9f/src/index.ts#L375-L377

to be eg.

```typescript
 const destFileName = path.resolve( 
   path.join(this.buildDirPath, `__only_${`${this.serverless.service.getServiceName()}-${ 
         this.options.stage || serverless.service.provider.stage 
       }-${functionAlias}`}`, filename) 
 ); 
 ```

File: src/index.ts [hunk 4]
```diff
@@ -365,15 +359,14 @@ export class EsbuildServerlessPlugin implements ServerlessPlugin {
     }
 
     // include any "extras" from the individual function "patterns" section
-    for (const fnName in this.functions) {
-      const fn = this.serverless.service.getFunction(fnName);
+    for (const [functionAlias, fn] of Object.entries(this.functions)) {
```
samchungy, line 361: Small refactor of existing code. `fnName` was confusing because it's actually the function's alias whereas fn.name is actually the name.

src/pack.ts
samchungy: Okay @floydspace I've replaced it with the `functionAlias` which is how the Serverless Framework packages zips so we should have no compatibility issues unless someone is doing something which the Serverless Framework does not support 😆.

I've also refactored a few confusing `fnName` variables to `functionAlias` around the codebase as well as changed `for ... in this.functions` with `for ... of Object.entries(this.functions)`

samchungy:
Hope you had a good new year @floydspace! Could you please give me your review again? Thanks!

floydspace (approved):

samchungy:
Purely for information sake - this PR should not force users to bump up their SLS to 2.50 version nor remove `serverless-pseudo-parameters`.

I've changed the approach since then. Just using the functionAlias instead of the function.name now which means we should have no more issues hopefully.

Issue #103, opened 2021-03-24
External dependency is not added to node_modules


ipauler:
I have an external dependency 'A' which is needed by other external dependency 'B'.
'A' is not used in my code so when function got bundled it doesn't adds 'A' into node_modules.
Is there a workaround for that ? Or maybe i can create a PR with a fix for that ?

olup:
If A is declared as external, it should be added to your node modules with all it's dependencies. If you only declare b as external it's indeed more complicated. What is your situation ? What did you declare in your external config ? And, for the record, what packages are your A and B ?

ipauler:
yes i'm doing individual packaging 
so the issue is with `util-deprecate` 
```
  "Runtime.ImportModuleError: Error: Cannot find module 'util-deprecate'",
        "Require stack:",
        "- /var/task/node_modules/readable-stream/lib/_stream_writable.js",
        "- /var/task/node_modules/readable-stream/readable.js",
        "- /var/task/node_modules/split2/index.js",
        "- /var/task/node_modules/pgpass/lib/helper.js",
        "- /var/task/node_modules/pgpass/lib/index.js",
        "- /var/task/node_modules/pg/lib/client.js",
        "- /var/task/node_modules/pg/lib/index.js",
        "- /var/task/src/app/graphql/handler.js",
 ```
`pg` is external in my case because it's not working other way

**Comment by olup** [comment]
Last question, are you in a mononrepo / workspace ? Are you bundling on you computer or in the cloud ? And in the former case, what os are you on ? Can you confirm util-deprecate is found in your local node_modules ?

**Comment by ipauler** [comment]
yarn workspaces
util-deprecate is in node_modules in root dir 
trying building it locally on mac and in CodeBuild with linux same result

**Comment by olup** [comment]
Thank you, I'll try to reproduce the error in the morning here.

**Comment by olup** [comment]
Would you mind sharing either a repo, either a copy-paste copy of your `package.json` (in your workspace and in the root) and your `serverless.yml`, so that I can reproduce a similar setup ?

**Comment by gusfune** [comment]
Hi, we are using a monorepo and the same issue is happening with us. In one function that we use `import Shopify from "shopify-api-node"` we get an error:

```
Error: Cannot find module './carrier-service'
Require stack:
- /orders/.esbuild/.build/src/orders/add_product.js
- /orders/node_modules/serverless-offline/dist/lambda/handler-runner/in-process-runner/InProcessRunner.js
[...]

PR #251, opened 2022-01-07
fix: make sure all the dependencies are checked when flattening the d…

danionescu:
…ependencies tree (#103)

This needs more testing. For example, my own project, I am getting the error `Error: Cannot find module 'source-map-support/register'`, which I have seen before my fix.

### DIFF
--- a/src/helper.ts
+++ b/src/helper.ts
@@ -94,7 +94,8 @@ const buildAllDeps = (deps?: JSONObject) => {
       Object.entries(dependencies).forEach((dep) => dependenciesLeft.push(dep));
     }
 
-    allDeps[depName] = details;
+    allDeps[depName] = allDeps[depName] || {};
+    allDeps[depName][(details as JSONObject).version] = details;
   }
 
   return allDeps;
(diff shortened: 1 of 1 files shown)


samchungy:
Would a package version affect the dependencies we need?

Eg. If one package had

X@0.1 as a dep

And then another had

X@0.2 as a dep

danionescu:
I am not sure how that would work with the current code. I pushed another commit that saves the dependencies for each version.

samchungy:
> I am not sure how that would work with the current code. I pushed another commit that saves the dependencies for each version.

Neither tbh, was just curious if you knew aha 😅

danionescu:
Lol, I took a bottom up approach to fixing this, assuming the rest of the is correct, so I just tried to add the pieces that were missing in this place. I would need to study the rest of the code to make sure it is correct, but I don't think that I will have the time soon.

samchungy (approved):
This looks good to me. Thanks for the contribution @danionescu!

I answered my above question with this helpful medium article. https://medium.com/learnwithrahul/understanding-npm-dependency-resolution-84a24180901b 

I did give this a run locally and it lgtm. It brings across the correct dependencies.

samchungy:
@olup @floydspace want to give this a final look?

floydspace (approved):

floydspace:
I pushed the button, thank you @danionescu @samchungy

PR #252, opened 2022-01-08
fix: add flat dependency tree

samchungy:
Idea for #103


samchungy:
#251 is doing the same thing

PR #261, opened 2022-02-04
fix: do not set package artefact for non-node functions

gridsam:
If you ask serverless to package your functions individually, the plugin will try to package everything, causing a crash if you have functions that were already filtered out by the code in `get functions()`.
(You can work around this by adding a dummy "- package: { artifact: '' }` property of the function.)

This just changes what functions need to be packaged to the ones that are valid.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -392,23 +392,12 @@ export class EsbuildServerlessPlugin implements ServerlessPlugin {
       path.join(this.serviceDirPath, SERVERLESS_FOLDER)
     );
 
-    if (this.options.function) {
-      const fn = service.getFunction(this.options.function);
-      fn.package.artifact = path.join(
-        this.serviceDirPath,
-        SERVERLESS_FOLDER,
-        path.basename(fn.package.artifact)
-      );
-      return;
-    }
-
-    if (service.package.individually) {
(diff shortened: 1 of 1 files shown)


samchungy:
> If you ask serverless to package your functions individually, the plugin will try to package everything, causing a crash if you have functions that were already filtered out by the code in `get functions()`. (You can work around this by adding a dummy "- package: { artifact: '' }` property of the function.)
> 
> This just changes what functions need to be packaged to the ones that are valid.

I think we must have missed this here: https://github.com/floydspace/serverless-esbuild/pull/141. Thank you!

src/index.ts
```diff
@@ -392,23 +392,12 @@ export class EsbuildServerlessPlugin implements ServerlessPlugin {
       path.join(this.serviceDirPath, SERVERLESS_FOLDER)
     );
 
-    if (this.options.function) {
```
samchungy, line 395: We set `this.functions` to only include this function and filter it out if it is not a node function.

We can reuse the same code block with package individually to ignore it if it is not a node function.

https://github.com/floydspace/serverless-esbuild/blob/573385c5642a31d243b55004997e7c513a67104d/src/index.ts#L170-L175

samchungy (approved):

PR #263, opened 2022-02-05
feat: enable esm support

### LINKED ISSUE #258: ESM Format Support
With [Lambda support for ESM](https://aws.amazon.com/about-aws/whats-new/2022/01/aws-lambda-es-modules-top-level-await-node-js-14/), it'd be awesome to be able to opt into ESM output format. Bonus points for being dynamic (since `serverless-offline` still only supports cjs, but prod bundles would want to be ESM)

samchungy:
Resolves #258

To enable ESM support simply override the `platform` option in the esbuild plugin settings. View the esbuild [documentation](https://esbuild.github.io/api/#platform) for more information. The `platform` key defaults to `node`. 

eg.
```yml
custom:
  esbuild:
    platform: neutral
```

## Change
- Allow `platform` to be overwritten in the options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ custom:
     minify: false
 ```
 
-Check [esbuild](https://github.com/evanw/esbuild#command-line-usage) documentation for the full list of available options. Note that some options like `entryPoints` or `outdir` cannot be overwritten. We rely on the Javascript API and so make sure you provide the JavaScript options as they can be different to the CLI options.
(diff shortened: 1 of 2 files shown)


samchungy:
I don't want to enable automatic serverless-offline `cjs` compilation as the difference between the two can vary quite drastically. If consumers wish to enable `serverless-offline` `cjs` compilation they could variably set the `platform` key.

eg.

You can achieve this in Serverless v3 very easily

eg.

```yml
params:
  default:
    platform: neutral
  test:
    platform: node

custom:
  esbuild:
    platform: ${param:platform}
```

floydspace (approved):
Makes total sense @samchungy , thank you.

PR #267, opened 2022-02-08
fix(external): update npm and yarn dependency detection

### LINKED ISSUE #103: External dependency is not added to node_modules

### LINKED ISSUE #135: External devDependencies are included into the bundle
I have uncovered some odd behaviour when bundling external dependencies. In our case it would always include dev dependencies into the build. I have dug a bit and found:

https://github.com/floydspace/serverless-esbuild/blob/master/src/pack-externals.ts#L78

which in our case would use the package.json of serverless-esbuild instead of the projects package.json. I am wondering if this should not actually be
```ts
  const packageJson = require(rootPackageJsonPath);
```

Because if

```ts
if (!packageJson.devDependencies || !packageJson.devDependencies[externalModule.external]) {
        prodModules.push(externalModule.external);
      } else {
```

refers to the package.json of serverless-esbuild it would add all dev dependencies, at least it does in our case.

Our setup looks like this:

```
node_modules
src
 - api
 -- serverless.yml

 - api2
 -- serverless.yml

 - api3
 -- serverless.yml

package.json
package-lock.json
```

### LINKED ISSUE #265: External dependencies bundling issue with version 1.23.4+
Hi

I have an issue after update from version 1.23.3 to 1.23.4
It seems like this update breaks external modules bundling

I have 1 external module:

```yml
    external:
      - sharp
```

This module has `color` dependency and `color` has `color-string` dependency. So I would expect to see both modules in the `/node_modules` folder. However with version 1.23.4 it seems like only top level dependencies are copied to final build:

v1.23.3 with external `sharp`:
image

v1.24.4+ with external `sharp`:
image

I think it can be related to this changes in 1.23.4: https://github.com/floydspace/serverless-esbuild/commit/91f283f587c3e005aa884dc528c82c3f82bca6c8 but I'm not sure about it

samchungy:
## Why
Resolves #265
Resolves #103
Resolves #135

- The current `yarn` dependency resolution logic does not take into consideration the version of the package it is looking for. This meant that if there was another version of a package of the same name with different dependencies we may not bring the correct dependencies into the package. This could cause extra or less packages to be included in the package.

- The current `npm` dependency resolution logic does not consider when a package is de-duped by another package. This leads to issues when multiple dependencies share the same version dependency.

## Change
- Unify output of the dependency tree shape and behavior of `npm` and `yarn` packagers. This should hopefully lead to more consistency.
- Update `flatdep` logic

## WIP Checklist
- [x] Come up with shape for unified dependency tree
- [x] Shape `yarn` dependency output to unified tree
- [x] Shape `npm` dependency output to unified tree
- [x] Shape `pnpm` dependency output to unified tree [Won't fix - see #269 ]
- [x] Change `flatDep` logic
- [x] Unit Test Coverage

### DIFF
--- a/src/helper.ts
+++ b/src/helper.ts
@@ -91,7 +91,7 @@ export const flatDep = (
       ]);
     }
 
-    // This is a nested dependency so we don't need to include it in the exclude
+    // This is a nested dependency so we don't need to include it in the bundle
     return uniq([...acc, ...flatDep(details.dependencies, undefined, rootDeps)]);
   }, []);
 };
--- a/src/helper.ts
+++ b/src/helper.ts
@@ -75,7 +75,7 @@ export const flatDep = (
   deps: DependencyMap | undefined,
   filter?: string[],
   rootDeps?: DependencyMap
-) => {
+): string[] => {
   if (!deps) return [];
 
   // keep tracks of the rootDeps when nested
@@ -83,8 +83,14 @@ export const flatDep = (
 
   return Object.entries(deps).reduce((acc, [depName, details]) => {
     if (filter && !filter.includes(depName)) return acc;
-    const dependencies = details.isRootDep ? rootDeps[depName].dependencies : details.dependencies;
-    return uniq([...acc, depName, ...flatDep(dependencies, undefined, rootDeps)]);
+    if (details.isRootDep || filter?.includes(depName)) {
+      return uniq([
+        ...acc,
+        depName,
+        ...flatDep(rootDeps[depName].dependencies, undefined, rootDeps),
+      ]);
+    }
+    return uniq([...acc, ...flatDep(details.dependencies, undefined, rootDeps)]);
   }, []);
 };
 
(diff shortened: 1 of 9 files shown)


lucduong:
@samchungy that's awesome, I'm looking for this PR. Did you finish all and only wait for merging and releasing?

samchungy:
> @samchungy that's awesome, I'm looking for this PR. Did you finish all and only wait for merging and releasing?

aha I finished last night at midnight my time, so I'm going to give it a day or two just to make sure I've got everything covered 😅

samchungy:
Feel free to give this a whirl. This one is a fairly big one so feel free to take your time.

To address the issues I published some packages and played around with their resolution.

I've given this a test with some real projects I have installed locally and it seems to work. Sorry for the large amount of PRs... 😅

Issue #277, opened 2022-02-22
Using this with serverless-offline pattern includes are only copied once


CorruptedHeart:
Given a package include like:
```
package:
  individually: true
  patterns:
    - "src/**/*.ejs"
```

The .ejs files will only be copied to output once, even if changes are made to the source files.

From the code it looks like it only copies if it doesn't already exist.

Was this for performance reasons?

Been thinking about how this could be fixed. Would the overhead of doing 2 stat calls per included file be worse than simply always copying? Doing the 2 stat calls would allow us to check that the file is newer before copying.

https://github.com/floydspace/serverless-esbuild/blob/master/src/index.ts#L324-L326

samchungy:
Yeah I think it was because of performance reasons, in particular if users are using copying in `node_modules` which can contain a ridiculous ammount of files.

samchungy:
> Yeah I think it was because of performance reasons, in particular if users are copying in `node_modules` which can contain a ridiculous ammount of files. Or if the user had big heavy artefacts

PR #278, opened 2022-02-22
Fix: Always copy pattern matched files

CorruptedHeart:
This allows offline development to always have the latest extra files in the output folder.

Switch to async file copying.

fixes: #277

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -321,9 +321,7 @@ class EsbuildServerlessPlugin implements ServerlessPlugin {
           fs.mkdirpSync(dirname);
         }
 
-        if (!fs.existsSync(destFileName)) {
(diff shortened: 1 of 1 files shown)


samchungy:
Change looks fine, but it won't fix your issue of not having the latest file. I believe we'd need to call `copyExtras` in the watch which I think might be what you are after?

https://github.com/floydspace/serverless-esbuild/blob/c30bf537d1d048945636013f8845164bbbc5f68a/src/index.ts#L196-L210

floydspace:
@CorruptedHeart I see there is more extended fix of the issue #277 in the PR #303 made by @tinchoz49
Could you review would it cover your change? Thank you.

CorruptedHeart:
Does appear to cover the use case. Happy to close this in favour of that.

PR #280, opened 2022-02-23
Add issue templates

samchungy:
We get alot of reports with not enough information and most of the time we have to ask them for logs, or steps to reproduce. 
These templates were generated automatically by GitHub. I just tweaked them to be more specific to our plugin


floydspace (approved):
@samchungy good to have it, thank you

Issue #259, opened 2022-02-01
Use Serverless V3 logging


fargito:
Hello ! Thanks for this awesome project :)

Serverless V3 came out last week and comes with a new logging system which produces cleaner output.
https://www.serverless.com/framework/docs/guides/plugins/cli-output

I think it would be great to use the same policy for this plugin, what do you think ? I can look into it and submit a PR in the next days.

floydspace:
Hi, @fargito thank you for the suggestion, I think it is a good idea, but keep in mind that the plugin can be still used with older versions of SLS, down to V1, not sure how these changes would affect the behavior, I will leave it to you. But PRs are welcome

samchungy:
I think this would force us to set a minimum version of serverless framework to v3 by the looks

samchungy:
Oh it looks like this is possible without a breaking change to previous versions

https://github.com/serverless-heaven/serverless-webpack/pull/1013/files

PR #282, opened 2022-02-24
Feat: Serverless v3 logs

fargito:
Hello!

I have implemented a small logging interface that makes use of the new logging capabilities of Serverless V3, but is still compatible with the legacy logger. I have tested with a project in V2 and another one in V3.

Cheers!

Fix: https://github.com/floydspace/serverless-esbuild/issues/259

### DIFF
--- a/src/pack-externals.ts
+++ b/src/pack-externals.ts
@@ -90,10 +90,9 @@ function getProdModules(
       !packageJson.dependencies[externalModule.external] &&
       !packageJson.devDependencies[externalModule.external]
     ) {
-      this.options.verbose &&
-        this.serverless.cli.log(
-          `INFO: Runtime dependency '${externalModule.external}' not found in dependencies or devDependencies. It has been excluded automatically.`
-        );
+      this.log.verbose(
+        `INFO: Runtime dependency '${externalModule.external}' not found in dependencies or devDependencies. It has been excluded automatically.`
+      );
 
       return;
     }
@@ -109,7 +108,7 @@ function getProdModules(
 
       if (!includes(externalModule.external, ignoredDevDependencies)) {
         // Runtime dependency found in devDependencies but not forcefully excluded
-        this.serverless.cli.log(
+        this.log.error(
           `ERROR: Runtime dependency '${externalModule.external}' found in devDependencies.`
         );
         throw new this.serverless.classes.Error(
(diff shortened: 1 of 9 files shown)


samchungy:
This is excellent 👏 🙇‍♂️. Love the interface over the approach the webpack PR took.

samchungy:
Interesting, so it seems that log.verbose = log.info in the v3 logging.

samchungy (approved):

src/pack-externals.ts
```diff
@@ -91,7 +91,7 @@ function getProdModules(
       !packageJson.devDependencies[externalModule.external]
     ) {
       this.options.verbose &&
```
samchungy, line 93: Could we include this in the `buildServerlessV3LoggerFromLegacyLogger` and remove these `this.options.verbose &&` statements?

```
export const buildServerlessV3LoggerFromLegacyLogger = (
  legacyLogger: (text: string) => void;
  verbose: boolean;
): ServerlessPlugin.Logging['log'] => ({
  error: legacyLogger,
  warning: legacyLogger,
  notice: legacyLogger,
  info: legacyLogger,
  debug: legacyLogger,
  verbose: verbose && legacyLogger || () => {},
  success: legacyLogger,
});
```
Something like this? What do you think?

fargito:
@samchungy actually this option is now natively supported by the framework so we can simply remove it

fargito:
@samchungy I've updated the PR :wink:

samchungy:
@fargito you legend! 🙌 I'll take a peek tomorrow! It's Friday drinks for me tonight😆🍻

floydspace (approved):
@fargito thank you for the contribution. approving it.

samchungy (approved):

src/pack-externals.ts
```diff
@@ -90,10 +90,9 @@ function getProdModules(
       !packageJson.dependencies[externalModule.external] &&
       !packageJson.devDependencies[externalModule.external]
     ) {
-      this.options.verbose &&
-        this.serverless.cli.log(
-          `INFO: Runtime dependency '${externalModule.external}' not found in dependencies or devDependencies. It has been excluded automatically.`
-        );
+      this.log.verbose(
```
samchungy, line 93: Hmmm 🤔  okay since `log.info = log.verbose` should we make these `verbose` logs now into `debug` logs?

so any `this.options.verbose && log.verbose` into the equivalent `this.log.debug`('this.options.debug && log.debug') ?

I'm happy to still go ahead with this though

PR #296, opened 2022-03-25
fix: use fs.createWriteStream instead of fs-extra

ArsenyYankovsky:
In one of our services the packaging process seems to be hanging [after opening the write stream to a zipped file](https://github.com/floydspace/serverless-esbuild/blob/a2516d754663a867416dc3bab200faf0863c3729/src/utils.ts#L140). The stream seems to be creates, but the `open` event never fires. Nothing is written to the file and no errors occur.

I couldn't find any difference between several projects and this one, and the only thing that seems to have helped is switching to using `createWriteStream` from the `fs` package instead of `fs-extra`. I don't think using `fs-extra` for this particular function provides any benefit? 

Weirdly enough, if I create a separate file like this and run it I get the contents written to the file:
```
const fs = require("fs-extra");
const output = fs.createWriteStream(/* zip path here */);

output.on('open', () => {
  output.write('test')
  output.end()
});
```

Please let me know how to procedd. Thanks.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,6 +2,7 @@ import { bestzip } from 'bestzip';
 import archiver from 'archiver';
 import childProcess from 'child_process';
 import fs from 'fs-extra';
+import nodeFs from 'fs';
 import path from 'path';
 import os from 'os';
 import { join } from 'ramda';
@@ -134,7 +135,7 @@ export const zip = async (
     fs.rmdirSync(tempDirPath, { recursive: true });
   } else {
     const zip = archiver.create('zip');
-    const output = fs.createWriteStream(zipPath);
+    const output = nodeFs.createWriteStream(zipPath);
 
     // write zip
     output.on('open', () => {


samchungy:
Hello what versions of serverless esbuild and nodejs have you got on your systems?

ArsenyYankovsky:
I tried with `serverless-esbuild` 1.23.4 and 1.26.1 (latest) and my node version is v16.5.0. Please note that the same combination of versions works for other projects.

samchungy:
I'm pretty sure fs-extra just creates a wrapper around the node fs one but if this works for you then sure, it shouldn't be too much of a biggie I guess

ArsenyYankovsky:
Yep, that's what I thought as well. It does some monkey-patching of streams though so that maybe a cause of the issues.
Strangely enough, I just tried the same project today and it worked. I guess I'll investigate deeper if it occurs again.

PR #301, opened 2022-04-04
Add variable to skip external packing

GioPat:
## Background
I have several serverless functions that share node_modules, I created a layer containing those node_modules so I do not need to package the dependencies inside the bundle.

### Additional information
I'm working on a windows machine and the tests and husky hooks are not properly working (paths in the expected unit test results are hardcoded in unix style). In future a PR that fixes this could be useful.

### DIFF
--- a/src/pack-externals.ts
+++ b/src/pack-externals.ts
@@ -204,7 +204,9 @@ function getProdModules(
  */
 export async function packExternalModules(this: EsbuildServerlessPlugin) {
   const plugins = this.plugins;
-
+  if (this.buildOptions.skipExternalsPacking) {
+    return;
+  }
   if (
     plugins &&
     plugins.map((plugin) => plugin.name).includes('node-externals') &&
--- a/src/types.ts
+++ b/src/types.ts
@@ -25,6 +25,7 @@ export interface Configuration extends Omit<BuildOptions, 'nativeZip' | 'watch' 
   keepOutputDirectory?: boolean;
   packagerOptions?: PackagerOptions;
   disableIncremental?: boolean;
+  skipExternalsPacking?: boolean;
 }
 
 export interface FunctionBuildResult {


samchungy:
Use exclude

GioPat:
If I put `./node_modules/*` in the exclude options esbuild is going to create a bundle that contains relative path (starting from the building directory) for the imports. This is not good since I need to keep the imports "untuched"

samchungy:
> If I put `./node_modules/*` in the exclude options esbuild is going to create a bundle that contains relative path (starting from the building directory) for the imports. This is not good since I need to keep the imports "untuched"


```
exclude

An array of dependencies to exclude from the Lambda. This is passed to the esbuild external option. Set to * to disable packaging node_modules
```

GioPat:
> > If I put `./node_modules/*` in the exclude options esbuild is going to create a bundle that contains relative path (starting from the building directory) for the imports. This is not good since I need to keep the imports "untuched"
> 
> ```
> exclude
> 
> An array of dependencies to exclude from the Lambda. This is passed to the esbuild external option. Set to * to disable packaging node_modules
> ```
I read it and I already tried it.

If I add `exclude: *`
I get a bundling error of a node_module dependency (sequelize)
image

samchungy:
Can you try setting external to `./node_modules/*` and exclude to `*`

GioPat:
> Can you try setting external to `./node_modules/*` and exclude to `*`

It's transpiling correctly but it's transforming the include using the relative paths.

samchungy:
> transforming the include using the relative paths.

Not quite sure what you mean by include using relative paths? Are you importing some node modules using relative paths?

GioPat:
> > transforming the include using the relative paths.
> 
> Not quite sure what you mean by include using relative paths? Are you importing some node modules using relative paths? Can you provide an example?

`import { MiddlewareObj } from "@middy/core";`
Gets transpiled into:
`var import_core = __toESM(require("../../../../../node_modules/@middy/core/index.js"));`
notice the path.

But by adding the nodeExternal plugin along with the two configs (exclude and external) seems to work fine

floydspace (commented):
Hi @GioPat 
Thanks for your contribution, but I don't think the change is necessary. Look at line 226, you'd see condition checking if the `exclude` option equal to `*` if it's equal to * the externals array would remain empty and packaging would not happen, it was implemented in the early stages exactly for the case like yours. Let me know if it serves your needs. Cheers!

GioPat:
Hello @floydspace thanks for the explanation and in fact using `exclude` set to `*` plus including esbuild-node-externals was successful to achieve my goal!

Thanks

PR #307, opened 2022-05-13
Fix `fs.rmdir` deprecation warning

dnicolson:
This pull request updates the `fs-extra` dependency and uses the `removeSync` method to avoid the following warning on Node.js v16+:

`[DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,13 +5,14 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "serverless-esbuild",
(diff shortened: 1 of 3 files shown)


floydspace (approved):
thank you @dnicolson 
makes total sense, approved

samchungy (approved):

PR #323, opened 2022-06-22
feat: add support for config file

tinchoz49:
Hey guys. 

This PR adds support to specify a config file. So the user can have more control about the configuration. 

For example: If I want to define all my dependencies as external I have to use a esbuild plugin or define in the serverless.yml the entire list of dependencies.

Using a config file lets the user to solve this by:

```js
module.exports = () => ({
  external: Object.keys(require('./package.json').dependencies)
})
```

Without having to install new dependencies or esbuild plugins that can impact in the performance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,26 @@ _Note:_ If you are using Python functions with Serverless Offline you will need 
 
 ## Advanced Configuration
 
+### Config file
+
+Esbuild configuration can be defined by a config file.
+
+```yml
+custom:
+  esbuild:
+    config: './esbuild.config.js'
+```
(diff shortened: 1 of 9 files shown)


floydspace (approved):
@tinchoz49 it looks good, thanks for your contribution.
The extra point here: you could also use https://github.com/serverless/typescript to define serverless config as TS file and do the same thing with importing `package.json` dependencies. However, it still requires having 181-184 lines from your PR. so I give a green light. 👍

tinchoz49:
Thank you @floydspace !

PR #336, opened 2022-07-16
fix: packagePath correctly treated as relative path from service dir

codingnuclei:
Hey.

This PR ensures that a provided `packagePath` is always resolved relative to the current working directory e.g serverless.yml dir.

Currently the `packagePath` is resolved based on the directory containing the currently executing file, i.e `/my-project/node_modules/serverless-esbuild/dist/pack-externals.js`;

https://github.com/floydspace/serverless-esbuild/blob/2c1d2ac842ac4270c48028d123887193ae957af2/src/pack-externals.ts#L78

If you change the `packagePath` with this in mind then it fails over when trying to resolve the `package-lock.json` later due to trying to resolve relative to the current working directory.

https://github.com/floydspace/serverless-esbuild/blob/2c1d2ac842ac4270c48028d123887193ae957af2/src/pack-externals.ts#L311

## Considerations:
- I have used `process.cwd()` to determine current working directory as seen in the util function `findUp`.
- This my have gone unnoticed as the log message for whether or not the lock file is used is LEVEL verbose.

### DIFF
--- a/src/pack-externals.ts
+++ b/src/pack-externals.ts
@@ -75,7 +75,7 @@ function getProdModules(
   packageJsonPath: string,
   rootPackageJsonPath: string
 ) {
-  const packageJson = require(packageJsonPath);
+  const packageJson = this.serverless.utils.readFileSync(packageJsonPath);
   const prodModules = [];
 
   // only process the module stated in dependencies section
@@ -132,6 +132,7 @@ function getProdModules(
       'package.json'
     );
     const localModulePackagePath = path.join(
+      process.cwd(),
       path.dirname(packageJsonPath),
       'node_modules',
       externalModule.external,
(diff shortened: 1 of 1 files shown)


floydspace (approved):
@codingnuclei, looks good to me. Thank you for contribution.

webdeveric:
I've noticed that `localModulePackagePath` now contains the `cwd()` twice since `path.dirname(packageJsonPath)` will begin with it already.

Example: **`/home/some-user/dev/example-project`**`/home/some-user/dev/example-project/node_modules/@middy/http-json-body-parser/package.json`

This fixes it:
```typescript
    const localModulePackagePath = path.join(
      process.cwd(),
      path.relative(process.cwd(), path.dirname(packageJsonPath)),
      'node_modules',
      externalModule.external,
      'package.json'
    );
```

codingnuclei:
> ```ts
> localModulePackagePath
> ```

Hey - I think in your case you don't need to set a `packagePath` at all as it would appear you are calling the command in the directory of the package.json.

Try removing the build option completely?

I think possible the docs could be improved to highlight that `packagePath` should be a relative path.

webdeveric:
@codingnuclei I do not have `packagePath` set in my config.

codingnuclei:
@webdeveric yes this does seem to be a bug introduced.

In the mean time, setting `packagePath: "./package.json"` should fix the problem in the immediate. 

I think fixing it higher up might be a better spot.
https://github.com/floydspace/serverless-esbuild/blob/f7f1e9b0d4cbf3a33351824bd429a394a69e949e/src/pack-externals.ts#L240-L241

`  const packageJsonPath =
    this.buildOptions.packagePath ||
    path.relative(process.cwd(), path.join(findUp('package.json'), './package.json'));
`

floydspace:
@all-contributors
please add @codingnuclei  for maintenance, ideas.
please add @webdeveric  for question, review.

PR #337, opened 2022-07-16
fix: set buildOptions in new initialize function

### LINKED ISSUE #334: esbuild config function is called with unresolved serverless param
**Describe the bug**
 Hi,

The first time `buildOptions` getter is called it is in the constructor.
https://github.com/floydspace/serverless-esbuild/blob/2c1d2ac842ac4270c48028d123887193ae957af2/src/index.ts#L85

If I am not wrong then this causes the esbuild config to be called with an unresolved serverless object.

> Note: configuration values are only resolved after plugins are initialized. Do not try to read configuration in the plugin constructor, as variables aren't resolved yet. Read configuration in lifecycle events only.
(https://www.serverless.com/framework/docs/guides/plugins/creating-plugins)

Possible moving this logic to the `initialize` hook might solve this.

**To Reproduce**
Steps to reproduce the behavior. If you have an example repository that would be even better:
1. Set : `esbuild:    config: "./esbuild.config.js"`
2. Set an ssm import somewhere in custom `SECRET: ${ssm.eu-west-1:/domain/${self:custom.stage}/secret}`
3. `console.log(serverless)` in the config function.
4.  `SECRET` will still be an un resolved string

**Expected behavior**
`SECRET` should be the resolved value from ssm

 - OS: [ Mac]
 - Serverless Framework Version: [2.72.3]
 - Plugin Version: [1.31.0]

codingnuclei:
closes #334

Hey,

This PR ensures that the buildOptions is set in the initialize hook. 
https://www.serverless.com/framework/docs/guides/plugins/creating-plugins#lifecycle-events

As per their docs the previous method meant that the serverless object passed to `config()` had variables that were not resolved...
> Note: configuration values are only resolved after plugins are initialized. Do not try to read configuration in the plugin constructor, as variables aren't resolved yet. Read configuration in lifecycle events only.

## Considerations:
- I have removed the dependency on `memoizeWith` as now the buildOptions is only set once.  As the class field has not changed name or the returning value, no changes are required in any other file.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 import fs from 'fs-extra';
 import globby from 'globby';
 import path from 'path';
-import { concat, always, memoizeWith, mergeRight } from 'ramda';
+import { concat, mergeRight } from 'ramda';
 import Serverless from 'serverless';
 import ServerlessPlugin from 'serverless/classes/Plugin';
 import chokidar from 'chokidar';
@@ -54,6 +54,7 @@ class EsbuildServerlessPlugin implements ServerlessPlugin {
   serverless: Serverless;
   options: Serverless.Options;
   hooks: ServerlessPlugin.Hooks;
+  buildOptions: Configuration;
   buildResults: FunctionBuildResult[];
   /** Used for storing previous esbuild build results so we can rebuild more efficiently */
   buildCache: Record<string, FileBuildResult>;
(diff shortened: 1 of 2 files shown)


floydspace:
@codingnuclei I will need more time to test and review it, so excuse me if it hangs longer. But shallow review shows me that the fix makes sense 😀

codingnuclei:
@floydspace no worries. Let's get this right! 👍

samchungy (approved):

samchungy (approved):
Gave it a run locally. Looks good!

Edweis:
Hey @codingnuclei , can we merge this PR? The fix will be appreciated. Thanks!

codingnuclei:
@Edweis hey - so as I understand it, I am unable to merge. 

Think @floydspace has to 😀

floydspace:
Merged it, thank you for contribution

Issue #315, opened 2022-05-30
Extra file (package.json) not included into final zips


doteric:
**Describe the bug**
Extra files (package.json to be exact) are not included into the final zips.

**Versions (please complete the following information):**
 - OS: Mac
 - Serverless Framework Version: 3.18.2
 - Plugin Version: 1.30.0

**Additional context**
So basically I wanted to use esbuild with ESM (AWS supports this for a while already). I set the `type` inside package.json to `module` and everything else needed for this move, worked fine, but the problem I stumbled upon is that serverless-esbuild simply does not include the `package.json` into the zip even if it's added to the `package.patterns` like so below.

```
custom:
  esbuild:
    bundle: true
    minify: false
    packager: 'yarn'
    target: 'node16'
    keepOutputDirectory: true

package:
  individually: true
  excludeDevDependencies: true
  patterns:
    - 'package.json'
```

It looks like the `.esbuild` contains the the `package.json`, but the zips themselves don't.

Of course I could just name my files `.mjs` instead, but that isn't an ideal solution for me.

floydspace:
hi @doteric , to be honest, not sure what is the reason we exclude, maybe because anyway we deliver bundled code and `package.json` used to be just unnecessary development artifacts, but feel free to test it locally with `patch-package` for instance and let as know, if you'd like raise a PR, keep in mind that we do not use root `package.json` for packaging dependencies, we generate one here https://github.com/floydspace/serverless-esbuild/blob/master/src/pack-externals.ts#L293, maybe it makes sense store this file, not the root one, but we also have to copy some necessary module meta located in root `package.json` you mentioned so your js files are treated correctly by nodejs.

doteric:
@floydspace , but this package.json seems to be generated only if there are external modules specified, right? Anyway it seems to be a bit mixed up. For now I'll hold with the PR as I don't have much time, but I assume the goal would be to use the base package.json and replace it with the necessary params (dependencies for example) generated by the package.json from the pack-externals.

floydspace:
@doteric yes, you're absolutely right, only external modules, here is the thing, other dependencies are bundled and you don't need them in generated package.json. so your assumption is correct, but could be vice versa, some of the params from base package.json would be copied to generated one, because you probably do not need `scripts` for instance nad other dev specific meta.

mpaleo:
@doteric Lambda relies on the package file or the file extension, so you could do something like this in your serverless.yml:
```
esbuild:
  format: esm
[...]

PR #339, opened 2022-07-18
fix: honour package.pattern over internally set exclusion list

codingnuclei:
Hey,

This PR enables the `package.pattern` to behave such that the user has the final say over what files are packaged.

Currently a list of predefined files are always excluded from the final zip, which is not great from a flexibility point of view #315 

https://github.com/floydspace/serverless-esbuild/blob/a233a1a5294ddf0875341312590f0078f468ae26/src/pack.ts#L38

Further more currently the files returned from the pattern glob are added to an include list...

https://github.com/floydspace/serverless-esbuild/blob/a233a1a5294ddf0875341312590f0078f468ae26/src/pack.ts#L171

... so although I may have used a glob like `!**/*.md`, those files will still be included into the final zip as the `filterFilesForZipPackage` will return true for any file that does not meet an exclusion criteria.

Taking the view that what is defined in both `package.pattern` and `function.pattern` as what files the user wants then we can remove files as needed.

Let me know what you think :)

## Considerations:
- Some perf improvements should be seen in larger repos as one use of `globby` has been removed.
- The glob search is ordered, so that a user can override the `excludedFilesDefault`.
- To keep this a none breaking change - the package.json is still being excluded by default. Users will manually have to include it via patterns. In the future this could be reversed.


OlivierCuyp:
Btw, I'm not event sure why we should exclude any files from the build folder.
All package.* files won't be there except if it is requested by the user...

paulgalow:
[Amazon Inspector can now scan Lambda functions for vulnerabilities](https://aws.amazon.com/de/blogs/aws/amazon-inspector-now-scans-aws-lambda-functions-for-vulnerabilities/). However [it looks like it is using `package.json` and `package-lock.json`](https://dev.to/rosswilliams/comment/238h9) to achieve this for Node.js functions. So having an option to manually include those files in the ZIP artifacts to be deployed would be really useful.

PR #342, opened 2022-07-19
fix: enable `esm` format

samchungy:
Recommend setting `format` to `esm` to enable ESM support. This PR enables us to recognise when we are in "ESM" mode or not.

Leaving the platform as `node` has its advantages.

eg. 

```
All [built-in node modules](https://nodejs.org/docs/latest/api/) such as fs are automatically marked as [external](https://esbuild.github.io/api/#external) so they don't cause errors when the bundler tries to bundle them.
```


floydspace (approved):
@samchungy looks good, thanks

PR #344, opened 2022-07-20
refactor: `getDepsFromBundle` now walks an AST

webdeveric:
I noticed `getDepsFromBundle()` was not finding dynamic imports using `import()`.

I've updated the code to use `acorn` and `acorn-walk` to find dependencies imported using `import 'XXX'`, `import('XXX')`, and `require('XXX')`.


samchungy:
Awesome - much better than our hacky regex 👍

samchungy (approved):

PR #352, opened 2022-07-26
refactor: asynchronously create packager instance

webdeveric:
Rewrote instantiating a `Packager` to be asynchronous.

- correctly call it so that `this` is the correct plugin
- Packager instance is memoized


floydspace (approved):
Hey @webdeveric Thank you for your contribution and test coverage. I like the approach.
Though I didn't expect that the previous implementation was returning an unnecessary promise, now it becomes reasonable.

PR #386, opened 2022-11-01
Support Linking External Dependencies

aidant:
Currently, when specifying external dependencies, either manually or through `esbuild-node-externals`, there is no capability to link external dependencies to the project. This functionality is critical for applications where active development includes creating and maintaining multiple external packages. 

Support for linking external dependencies has been added through a new configuration property called `link`, which will accept an array of packages and, using the specified `packager` link the packages into the `node_modules` folder nestled in the `.esbuild` directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ Serverless plugin for zero-config JavaScript and TypeScript code bundling using 
 - [Advanced Configuration](#advanced-configuration)
   - [Including Extra Files](#including-extra-files)
   - [External Dependencies](#external-dependencies)
+  - [Linking External Dependencies](#linking-external-dependencies)
   - [Esbuild Plugins](#esbuild-plugins)
 - [Usage](#usage)
   - [Automatic Compilation](#automatic-compilation)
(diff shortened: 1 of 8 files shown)


floydspace (commented):
Hi @aidant, thank you for the interesting solution. Could you please give an example of what you are linking to nested `node_modules`? Do you have to initiate a link first and then run `serverless offline`? and why do you need to do linking during deployment?
it would be helpful if you create an exemple in `examples` folder

aidant:
Hey @floydspace,

I can create an example in the `examples` folder later today or tomorrow. For now, I can explain our scenario: 

At @agriwebb we share several utility packages between our serverless projects. Before we introduced `serverless-esbuild`, our workflow would involve linking these packages into our serverless projects; this linking would allow us to develop and test these packages locally without publishing.

For unrelated reasons, in our `serverless-esbuild` projects, we have opted to use `esbuild-node-externals`. As you know, this causes the `dependencies` to be installed by the respective package manager in `./.esbuild/.build/node_modules`. The problem we encounter here is when linking the packages into the project, those linked packages are symlinked to `./node_modules` and are therefore not used in the serverless bundle when starting `serverless offline`.

This feature would allow us to run `npm link` in a utility package's workspace - let's call it `@agriwebb/utils`, then test the package locally in a serverless project by adding `@agriwebb/utils` to the `link` array.

We tried linking the packages to `./.esbuild/.build/node_modules`; however, in our testing, this gets overridden when running `serverless offline`.

We currently don't need linking during deployment; however, when testing, infrastructure-related changes have previously deployed to playground environments from our local machines.

PR #387, opened 2022-11-01
feat: allows .jsx and .tsx as valid handler extensions

bsdahl:
### This PR adds `.jsx` and `.tsx` as valid extensions to the `extractFunctionEntries` helper method.

My organization uses .jsx/.tsx serverless handlers to generate and send emails using react-dom/server renderToStaticMarkup. However, serverless-esbuild cannot find our handlers, because it is only checking for `.js` and `.ts` files.


floydspace (approved):

floydspace:
@bsdahl it makes total sense in this scenario, thank you for contribution

PR #392, opened 2022-11-19
fix: node18 support

huksley:
🎉  https://aws.amazon.com/blogs/compute/node-js-18-x-runtime-now-available-in-aws-lambda/


floydspace (approved):

floydspace:
Thank you @huksley

PR #393, opened 2022-11-26
refactor: node externals

webdeveric:
- Added `esbuild-node-externals` as a dev dependency and optional peer dependency
  - This allows us to use the types from their package during development.
- Refactored how `esbuild-node-externals/dist/utils` was being detected.
  - `require.resolve()` is now used instead of a hard coded relative path.


floydspace (approved):
@webdeveric LGTM, thanks for the improvement.

PR #394, opened 2022-12-02
Look up root workspace package.json for dependencies when using externals

marc-plaza:
The current logic of the plugin when using the `external:` option is to look if the dependencies specified in it are also in package.json `dependencies` or `devDependencies` and to ignore if not found there.

However in a monorepo setup, it is common to have shared dependencies in the root package.json and these might be wanted to be in external for whatever reason (in our use-case, we have a monorepo with lambdas sharing common dependencies such as `aws-sdk`, that need to be required instead of being bundled for calls intercepting by OpenTelemetry/Dynatrace)

This PR updates `getProdModules` to also lookup the root `package.json` in addition to the package `package.json` when looking for dependencies using `externals:`

This fix  solved the issue in our setup, but we might be have not enough insight for all use-cases, we will happily update this MR if some additional logic is needed.

### DIFF
--- a/src/pack-externals.ts
+++ b/src/pack-externals.ts
@@ -274,9 +274,15 @@ export async function packExternalModules(this: EsbuildServerlessPlugin) {
 
   // (1) Generate dependency composition
   const externalModules = map((external) => ({ external }), externals);
-  const compositeModules: JSONObject = uniq(
+  let compositeModules: JSONObject = uniq(
     getProdModules.call(this, externalModules, packageJsonPath, rootPackageJsonPath)
   );
+
+  if (isWorkspace) {
+    compositeModules = compositeModules.concat(
+      getProdModules.call(this, externalModules, rootPackageJsonPath, rootPackageJsonPath)
+    );
+  }
 
   if (isEmpty(compositeModules)) {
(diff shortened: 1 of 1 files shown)


floydspace:
Hi @marc-plaza 
Thanks for your contribution. Not sure if it would impact someone, but the tests look good.
For better confidence, I would suggest writing an e2e test similar to what we have in the examples folder.
But I'm fine to take a risk and merge it as is because not many changes are done.
Let me know if you have an idea of how to write the test, otherwise, I just merge it.

floydspace:
BTW @marc-plaza shouldn't dependencies always be described in the package level file? if some dep needs to be packed why it's searched in the root?

marc-plaza:
> BTW @marc-plaza shouldn't dependencies always be described in the package level file? if some dep needs to be packed why it's searched in the root?

Because we have an increasing number of  "packages" (lambdas), each one of them using the exact same dependencies hence why the dependencies are in root and not in the packages themselves, it would be redundant information and tedious to update if the deps were copy/pasted in each package.

As for the test, I don't really know how to implement one for that use-case given your examples, but I can look into it and try.

By the way, happy new year !

PR #404, opened 2022-12-09
chore: re-introduce node v14 compatibiliy, bump deps

dnalborczyk:
we have some scenario tests running in the `serverless-offline` repository with `serverless-esbuild`. bumping esbuild to `v0.15.13+` results into the tests hanging. this was fixed in 0.16.3: https://github.com/evanw/esbuild/issues/2727

unfortunately `esbuild-node-externals` has a peer dependency on 0.15 as well, which ideally would need to be fixed prior: https://github.com/pradel/esbuild-node-externals/pull/35

we can wait and introduce **esbuild v0.16** compat in a follow-up PR.

- re-introduces node v14 compatibility
- compiler target to node v14 supported: ES2020
- bumps all (cjs) deps


floydspace (approved):

floydspace:
@dnalborczyk thank you for your contribution, it makes sense. I made fixes in the e2e tests and ci/cd required to make it work.

dnalborczyk:
thank you for picking this up @floydspace !

PR #408, opened 2022-12-17
chore: `tsconfig.json` updates

webdeveric:
- Updated TS config
  - enabled `strict` mode
  - added extra config for building `src` only
  - updated type checking
- Updated helpers/utils
  - added type predicate and assertion functions
- Updated types due to `strict: true`


floydspace:
I gonna look at the PR more, but so far I noticed that you made target es2021, apparently for node 16, 
today I merged the PR where the node14 runtime came back, so having es2020 makes more sense.

floydspace (approved):
@webdeveric, LGTM thank you.

PR #413, opened 2022-12-27
refactor utils with fp-ts, cover zip function with unit tests

floydspace:
Would love to migrate everything to FP. Raising this PR as a starting point for using `fp-ts`


src/utils.ts
```diff
@@ -45,45 +58,41 @@ export function spawnProcess(command: string, args: string[], options: childProc
...
  * Forwards `rootDir` or finds project root folder.
  */
-export function findProjectRoot(rootDir?: string): string | undefined {
-  return rootDir ?? findUp(['yarn.lock', 'package-lock.json']);
-}
+export const findProjectRoot = (rootDir?: string) =>
+  pipe(
+    IOO.fromNullable(rootDir),
+    IOO.fold(() => findUpIO('yarn.lock'), IOO.of),
+    IOO.fold(() => findUpIO('package-lock.json'), IOO.of),
```
webdeveric, line 93: Should `pnpm-lock.yaml` be here too? I know support for pnpm isn't great at the moment.

webdeveric (approved):

src/utils.ts
floydspace: seems yes, thanks.

PR #414, opened 2022-12-28
feat: add support for dir/index handlers

jeffmarshall:
Adds code (and a test) that searches for the handler file as an `index.[js|ts|tsx|jsx]`. This way one can point to a directory with an index file that exports the handler function, instead of requiring the author to name the exact file that exports the handler.

Directory structure:

```
- handlers
  - foo
    - index.ts
```

`serverless.yml`:

```
# ...
functions:
  foo:
    handler: handlers/foo.foo
```

### DIFF
--- a/src/helper.ts
+++ b/src/helper.ts
@@ -77,6 +77,13 @@ export function extractFunctionEntries(
           func,
           functionAlias,
         };
+      } else if (fs.existsSync(path.join(cwd, path.join(fileName, 'index') + extension))) {
+        const entry = path.relative(cwd, path.join(fileName, 'index') + extension);
+        return {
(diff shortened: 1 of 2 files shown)


floydspace (approved):
Hi @jeffmarshall thank you for the contribution , could you please mention this feature in readme, so it wouldn't be hidden. Anything else lgtm

jeffmarshall:
> Hi @jeffmarshall thank you for the contribution , could you please mention this feature in readme, so it wouldn't be hidden. Anything else lgtm

I'm not sure exactly what to say in the Readme. I think most people would assume this behaviour is supported. All I changed was the step that finds the files to watch for changes. Maybe I should have called this change a "fix"?

floydspace:
@jeffmarshall okay I will make it as a fix when merging, no worries, thanks again

PR #421, opened 2023-01-04
feat: adds `noInstall` option for Yarn packager

chrishutchinson:
I've been trying to get `serverless-esbuild` working with my Yarn 3.x setup, and as far as I understand the package is not compatible at the moment. After doing some testing locally, and comparing this to `serverless-webpack`, it looks like just deactivating the install step should be enough to get it working in my context.

As such, and following the implementation in `serverless-webpack` [here](https://github.com/serverless-heaven/serverless-webpack/pull/1003), I have added a `noInstall` `packagerOptions` option, which exits out of the install process early. This trusts I have installed the dependencies myself.

I wanted to add an e2e test to cover this scenario but I couldn't quite follow the setup of `individually` vs. `complete` etc. I think I actually just want a dedicated e2e suite that is a copy of minimal with Yarn as the package manager, but I'd happily take guidance on what that should look like.

Thank you!


floydspace:
hi @chrishutchinson  not sure how to cover it with e2e test either. seems like `actions/setup-node` action does not support it yet.
anyway the code looks good to me, it does not violate existing functionality and I'm ready to give it a try

src/packagers/yarn.ts
```diff
@@ -39,6 +39,14 @@ const getNameAndVersion = (name: string): { name: string; version: string } => {
  *   ignoreScripts (false) - Do not execute scripts during install
  */
 export class Yarn implements Packager {
+  private packagerOptions: PackagerOptions;
+
+  constructor(packagerOptions?: PackagerOptions) {
+    this.packagerOptions = packagerOptions || {
```
floydspace, line 45: What I would improve here is having this default option in the root configuration defaults
Here: https://github.com/floydspace/serverless-esbuild/blob/ec4aab91af3b50f32adb791be9e5399ca0c57228/src/index.ts#L279

floydspace (approved):

src/packagers/yarn.ts
chrishutchinson: @floydspace Ah yeah, I had hesitated about doing this as it is only compatible with Yarn right now, as passing it for NPM wouldn't do anything. Instead setting the default in the `yarn.ts` file felt more _correct_. Happy to take your steer on it though, I can add it in here if you'd prefer

floydspace: --noInstall sounds generic. So potentially can be used for all packages.

chrishutchinson: Yeah you're right – I guess I was _just_ doing the Yarn implementation in this PR to keep it small, assuming it could be easily extended by others as-needed.

I will make an additional commit that includes it in this list of defaults just so it's clear. At least a default of `false` is the standard existing functionality for all packagers

chrishutchinson: Done in https://github.com/floydspace/serverless-esbuild/pull/421/commits/4b527a9497c43e9cf88796479bd3a9cc0d8825ca

floydspace (approved):
@chrishutchinson lgtm, thank you for your contribution👍🏼

floydspace:
@all-contributors please add @chrishutchinson for the new code, small but valuable contribution

Issue #376, opened 2022-09-28
Unable to package and deploy from an s3 bucket


brandonin:
**Describe the bug**
When adding the serverless-esbuild plugin, it always builds and does not take into account s3 buckets that are added as artifacts. It overrides any artifacts that are provided in favor of the location that esbuild outputs to.

https://github.com/serverless/serverless/issues/11414

**To Reproduce**
```
package:
  individually: true

plugins:
  - serverless-esbuild

provider:
  name: aws
  runtime: nodejs14.x
  region: us-west-2

functions:
  addData:
    handler: addData.main
    package:
      artifact: s3://some-bucket/add-data.zip
    events:
      - http:
          path: add-data/{code}
          method: post
```

**Expected behavior**
It should deploy from an s3 artifact.

**Versions (please complete the following information):**

Mac OSX 12.2.1 (21D62)
Framework Core: 3.22.0 (local) 3.22.0 (global)
Plugin: 6.2.2
SDK: 4.3.2

koryhutchison:
I'm experiencing the same issue. It appears that esbuild is trying to compile it but really it shouldn't be doing anything if an artifact is being provided. In my situation I've got a project that has several lambas and leverages esbuild, but I'm also adding another lambda to the project that's going to use an artifact that's published to S3. So for that lambda I don't want esbuild to do anything with it.

koryhutchison:
So I've done some investigation and testing and this isn't just limited to S3. Basically if any function in the serverless config has an `artifact` defined, then the plugin throws an error because it can't find the handler. When really, if an artifact is defined, esbuild doesn't need to run at all. It's already zipped up. So this includes local `artifact`s as well as ones in S3.

I think I have an easy fix for this, but I want to know if you think it's acceptable or not. The problem stems from [here](https://github.com/floydspace/serverless-esbuild/blob/master/src/index.ts#L149) because `functions` just picks up on all the functions in the config. (Except for those that aren't Node functions) So my suggested change would be to add a build option called `excludeFunctions` or something like that where I could use that to tell serverless-esbuild to just ignore those particular functions all together. So an example yaml would be like:

```yaml
custom:
  esbuild:
    excludeFunctions:
      - functionAlias
```

And then the extra check would be to just check if the function has been excluded:

```
private isExcluded(functionAlias: string): boolean {
    return this.buildOptions.excludeFunctions.includes(functionAlias);
}
```

[...]

PR #423, opened 2023-01-05
feat: add skipEsbuild option on a per function level

koryhutchison:
### Use case:

I've got multiple projects where I'm using serverless-esbuild and it's working great. But I'm building my own custom plugin for these projects where I'm adding a function and defining my own artifact zip file. Turns out though, serverless-esbuild will try to run esbuild even if an artifact is already defined. I'm sure this use case wasn't something that was thought of in the beginning, but I feel like it's something this should account for. 

This has also been brought up in #376

### Solution:

I tried going a variety of different routes, but the best solution I have so far is to make it so you can set a property when you define the function called `skipEsbuild` and if it's `true`, then it'll just skip over it. I'm leveraging the [Serverless schema configuration helpers](https://www.serverless.com/framework/docs/guides/plugins/custom-configuration) so that some default validation takes place.

I'm still pretty new to TypeScript and all, so if I did something incorrect here, let me know!


floydspace (approved):
hi @koryhutchison thanks, look good to me. I think it will cover a lot of cases. People were asking for this

floydspace:
@all-contributors please add @koryhutchison for new code and idea

PR #426, opened 2023-01-09
feat: support installing with Yarn berry flags

chrishutchinson:
In addition to the `noInstall` flagged added in PR #421, I have an additional use case where I still need to let `serverless-esbuild` proceed with the install in a Yarn 3 app. As such, I have brought over the Yarn Berry (>= 2) install method from `serverless-webpack`, as appropriate for `serverless-esbuild`. In effect it conditionally adds the appropriate install flags depending on the existing `useLockfile` flag, and the Yarn version (determined by two new methods).

_The implementation here is copied almost wholesale from the `serverless-webpack` implementation, here: https://github.com/serverless-heaven/serverless-webpack/blob/master/lib/packagers/yarn.js#L139_

### DIFF
--- a/src/packagers/yarn.ts
+++ b/src/packagers/yarn.ts
@@ -230,13 +230,25 @@ export class Yarn implements Packager {
       return;
     }
 
+    const version = await this.getVersion(cwd);
+    const isBerry = this.isBerryVersion(version);
+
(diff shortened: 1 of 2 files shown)


chrishutchinson:
Closing temporarily, as this implementation causes some issues in my testing around the `yarn.lock` file. I'll re-open it once I've investigated further and found the right approach.

Issue #375, opened 2022-09-27
Patterns work differently from node 14 to node 16


clementAC:
**Describe the bug**
I have a configuration that works on node 14 and as soon as I upgrade to node 16, the configuration did not work. 
I'm trying to delete some file from @prisma deps as described in their [docs](https://www.prisma.io/docs/guides/deployment/deployment-guides/deploying-to-aws-lambda#package-pattern-in-serverlessyml).
The deletion works on .prisma and prisma directory but on node 16 I still have the content of @prisma/engines whereas it was not the case in node 14.

here is what I tried so far:

```yaml
  esbuild:
    bundle: true
    minify: false
    sourcemap: true
    packager: npm
    installExtraArgs: ['--production']
    exclude: ['aws-sdk', 'pg-native']
    external: ['pg']
    plugins: plugins.js
    keepOutputDirectory: true
.......
package:
  patterns:
    - '!node_modules/.prisma/**'
    - '!node_modules/prisma/libquery_engine-*'
    - '!node_modules/@prisma/engines/**'

  individually: true

functions:
  graphql:
    timeout: 30
    memorySize: 512
    handler: src/lambdas/graphqlLambda.handler
    package:
      patterns:
        - 'node_modules/.prisma/client/libquery_engine-rhel-*'
        - 'node_modules/.prisma/client/*.js'
        - 'node_modules/.prisma/client/schema.prisma'
    events:
......
```
I'm only adding prisma to function that need it on function pattern level

**To Reproduce**

Quite complicated to reproduce as prisma need a db connection.

**Expected behavior**
To have patterns directives fully taken into account

**Versions :**

- OS: MacOS 12
- Serverless Framework Version: 3.22.0
- Plugin Version: 1.33.0
- esbuild-node-externals: 1.5.0

**Additional context**
Prisma 4

tstackhouse:
I'm seeing similar weird behaviour of the `patterns` option.  I'm specifying patterns that are in my dist folder as part of other steps of the build process, but as this plugin hook into/overrides parts of the package step, those patterns don't seem to be respected, or at least something to do with the cwd that globby is ran in to calculate include/excludes

PR #428, opened 2023-01-17
feat: handle exclude patterns for functions

tstackhouse:
Followup to my comment on #375 

By explicitly looking at the build dir and factoring out exclude patterns from the function level, we can better handle individual packaging of functions, especially when other build tooling is in play.


floydspace (approved):
Hi @tstackhouse, thank you for the fix, looks good to me

PR #429, opened 2023-01-20
fix: node matcher for google

descampsk:
Hello,

The current implementation of NodeMatcher fails when using with Google Cloud because Google expects a different format for the runtime.

Aws expects a format like `nodejs${Version}.x`
Google expects a format like `nodejs${Version}`

This PR allows the use of both format depending of the provider.


floydspace (approved):
@descampsk I thought no one use the plug-in with gcp, thank you for fixing it.

PR #431, opened 2023-01-26
fix: remove ! from exclusion pattern list for proper matching

tstackhouse:
My last PR (#428) had a small boolean logic error, by grabbing all the exclude patterns and then simply running globby on them, we were effectively gathering nothing, rendering the change ineffectual.  In order to properly gather the list of files to exclude, we need to remove the `!` from the exclusion patterns so we can actually get the results we want from globby.


floydspace (approved):
thank you @tstackhouse , looks good

Issue #433, opened 2023-02-06
1.35.3 breaks when using yarn 3


dhmw:
**Describe the bug**

Updating to 1.35.3 does not work with yarn 3. Specifically, I am using 3.1.1.

1.35.2 works as expected.

**To Reproduce**

Use serverless-esbuild 1.35.3 or greater with a project using yarn 3.1.1

**Expected behavior**

Build completes successfully

**Screenshots or Logs**

```
Packaging [REDACTED] for stage test (eu-west-1)
Warning: No external modules needed
Environment: linux, node 18.14.0, framework 3.27.0 (local), plugin 6.2.3, SDK 4.3.2
Credentials: Local, environment variables
Docs:        docs.serverless.com
Support:     forum.serverless.com
Bugs:        github.com/serverless/serverless/issues

Error:
Error: Command failed with exit code 1: yarn list --json --production
Usage Error: Couldn't find a script named "list".

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] <scriptName> ...
    at makeError (/home/[REDACTED]/node_modules/execa/lib/error.js:60:11)
    at handlePromise (/home/[REDACTED]/node_modules/execa/index.js:118:26)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Yarn.getProdDependencies (/home/[REDACTED]/node_modules/serverless-esbuild/dist/packagers/yarn.js:39:35)
    at async EsbuildServerlessPlugin.pack (/home/[REDACTED]/node_modules/serverless-esbuild/dist/pack.js:116:53)
    at async before:package:createDeploymentArtifacts (/home/[REDACTED]/node_modules/serverless-esbuild/dist/index.js:73:17)
    at async PluginManager.runHooks (/home/[REDACTED]/node_modules/serverless/lib/classes/plugin-manager.js:530:9)
    at async PluginManager.invoke (/home/[REDACTED]/node_modules/serverless/lib/classes/plugin-manager.js:563:9)
    at async PluginManager.run (/home/[REDACTED]/node_modules/serverless/lib/classes/plugin-manager.js:604:7)
    at async Serverless.run (/home/[REDACTED]/node_modules/serverless/lib/serverless.js:179:5)
    at async /home/[REDACTED]/node_modules/serverless/scripts/serverless.js:812:9
```

**Versions (please complete the following information):**

[...]

PR #436, opened 2023-02-15
fix: external incompatibility with Yarn 2+; conditionally set packager args

### LINKED ISSUE #435: serverless-esbuild doesn't work with `external` in Yarn 3+
**Describe the bug**
If I configure an `external` option in the serverless-esbuild custom config (or try using the esbuild-node-externals plugin), when trying to run Serverless offline mode or Serverless deploy, I get an error that prevents running offline mode or deploying the Lambda. Based on the error message it seems specific to Yarn 3+.

**To Reproduce**
Steps to reproduce the behavior.
Example repo: https://github.com/track0x1/sls-esbuild-bug

1. Clone the repo
2. `yarn` to install dependencies
3. `yarn start`
4. Notice the error and the fact that offline mode does not start up.

**Expected behavior**
The commands function as expected. (Deploy or run in offline mode)

**Screenshots or Logs**
```
Error:
Error: Command failed with exit code 1: yarn install --frozen-lockfile --non-interactive
➤ YN0050: The --non-interactive option is deprecated
```

- OS: Mac
- Node: 18.13.0
- Serverless Framework Version: 3.27.0
- Plugin Version: 6.2.3
- Yarn version: 3.3.1

**Additional context**
N/a

track0x1:
As a result of my issue filed in #435, we can detect the version of Yarn being used by the user and set the appropriate packager args.
* `--non-interactive` was removed from Yarn 2+
* `--frozen-lockfile` was changed to `--immutable` in Yarn 2+
* However, because of the new nature of Yarn `--immutable`, if you are using `external` dependencies the lockfile will always fail because we generate a new package.json with only the externalized dependencies. Therefore, I added `ignoreLockfile` as a Yarn only packagerOption to bypass this.

Closes #435 #433

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,10 +105,11 @@ The following `esbuild` options are automatically set.
 
 #### Packager Options
 
-| Option    | Description                                                                                           | Default     |
-| --------- | ----------------------------------------------------------------------------------------------------- | ----------- |
-| `scripts` | A string or array of scripts to be executed, currently only supports 'scripts' for npm, pnpm and yarn | `undefined` |
-| `noInstall` | [Yarn only] A boolean that deactivates the install step | `false` |
+| Option           | Description                                                                                                                                                                        | Default     |
+| ---------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------- |
+| `scripts`        | A string or array of scripts to be executed, currently only supports 'scripts' for npm, pnpm and yarn                                                                              | `undefined` |
+| `noInstall`      | [Yarn only] A boolean that deactivates the install step                                                                                                                            | `false`     |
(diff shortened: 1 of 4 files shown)


lukevers:
@track0x1 this is great, thanks. did bombas finally switch from yarn 1 to berry??

just tested and this works for me:

```
    "serverless-esbuild": "track0x1/serverless-esbuild#fix-yarn-compatibility",
```

track0x1:
> @track0x1 this is great, thanks. did bombas finally switch from yarn 1 to berry??
> 
> just tested and this works for me:
> 
> ```
>     "serverless-esbuild": "track0x1/serverless-esbuild#fix-yarn-compatibility",
> ```

Thanks! I was using this on a personal project and wanted to get it working with Berry; we're still on v1, though I would like to upgrade to enjoy some of those performance benefits!

track0x1:
@floydspace Would you mind taking a look at this PR? 🙏

floydspace (approved):
Hi @track0x1, looks good

PR #438, opened 2023-02-20
Allow overriding chokidar options

### LINKED ISSUE #290: Slow detection of file changes
I see that when I use serverless-offline, after I write a file it takes 2-3 seconds until the new code is compiled. If I try to run a function in this time, the old code is run. This is very inconvenient because it means I need to watch the output of serverless-esbuild after I make changes and wait until I see that the compile has completed before I run the function again.

I checked out what is causing this and I see that it's the `awaitWriteFinish` option for chokidar. Is this really necessary to use? I can't recall ever having an issue with half-written files, but I suppose it may vary between machines.

I propose two fixes for this:
1. When a file is changed block requests to functions until the compilation is finished. With this I would still have to wait 2-3 seconds, but at least I wouldn't have to watch the output of esbuild and wait with performing my request.
2. Make the `awaitWriteFinish` option (and perhaps the other chokidar options) configurable.

Preferably, both of these fixes would be implemented.

Could you consider this, or would you accept a PR for this?

trygveaa:
This adds support for overriding chokidar options by setting them under the watch section of the options. This is useful because the appropriate value for `awaitWriteFinish` is dependent on the OS and hardware.

I also removed setting `awaitWriteFinish` by default as I don't think it's a good idea as I explain in the commit message. Though, if you disagree with this, I'd rather get the rest of the PR merged without that than nothing at all, so I at least can unset it myself.

I also changed the options to be deeply merged, because otherwise you have to set all the default values under a section if you set a value in it. E.g. if you set anything under watch, you had to specify patterns as well, otherwise nothing is watched. With a deep merge, only the specific options you set are overridden.

Fixes #290

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,10 +112,10 @@ The following `esbuild` options are automatically set.
 
 #### Watch Options
 
-| Option    | Description                                                                                          | Default                                                |
-| --------- | ---------------------------------------------------------------------------------------------------- | ------------------------------------------------------ |
-| `pattern` | An [anymatch-compatible definition](https://github.com/es128/anymatch) for the watcher to respond to | `./\*_/_.(js\|ts)` (watches all `.js` and `.ts` files) |
-| `ignore`  | An [anymatch-compatible definition](https://github.com/es128/anymatch) for the watcher to ignore     | `'.esbuild', 'dist', 'node_modules', '.build']`        |
+| Option    | Description                                                                                          | Default                                               |
+| --------- | ---------------------------------------------------------------------------------------------------- | ----------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


floydspace (approved):
Hi @trygveaa thanks for good explanation of the fix and good PR quality overall . Looks like valuable fix, though Ive never used "watch" with this plugin.

trygveaa:
> Hi @trygveaa thanks for good explanation of the fix and good PR quality overall.

Great, thanks for the fast merge!

> Looks like valuable fix, though Ive never used "watch" with this plugin.

Oh, how do you do local development then? I use it together with serverless-offline when developing on my local machine.

floydspace:
> Oh, how do you do local development then? I use it together with serverless-offline when developing on my local machine.

Yes with sls-offline, but without watch. Besides, I already long time prefer using AWS CDK

PR #440, opened 2023-02-23
fix: don't dispose build contexts

### LINKED ISSUE #439: Getting "✖ Bundle error, waiting for a file change to reload..." with esbuild 0.17.10
**Describe the bug**
"✖ Bundle error, waiting for a file change to reload..." error keeps showing with esbuild 0.17.10 when file changes but doesn't show on first start. New changes are not applied.

**To Reproduce**
1. Start serverless-offline
2. make changes

Reproducible repo https://github.com/deathemperor/serverless-reprodution

**Expected behavior**
No error showns

**Screenshots or Logs**
image

- OS: Mac
- Serverless Framework Version: 3.27.0
- Plugin Version: 1.39.0

**Additional context**
Doesn't occur with esbuild 0.16.x

trygveaa:
Previously the build results were stored in a cache, but the context was immediately disposed. This caused an error because the context object was still in the cache, so when rebuilding it tried to call .rebuild() on the context on line 97 which threw an exception because you can't call rebuild on disposed contexts.

I don't get why contexts are stored in a cache and tried rebuilding on if they are disposed, so I removed disposing them all together. If they should still be disposed, the context needs to be completely removed, so it doesn't try calling rebuild on it.

Fixes #439

### DIFF
--- a/src/bundle.ts
+++ b/src/bundle.ts
@@ -161,10 +161,5 @@ export async function bundle(this: EsbuildServerlessPlugin, incremental = false)
     })
     .filter((result): result is FunctionBuildResult => typeof result === 'object');
 
-  // dispose of long-running build contexts
-  Object.entries(this.buildCache).forEach((entry) => {
-    entry[1].context?.dispose();
-  });
-
   this.log.verbose('Compiling completed.');
 }


jlangsu:
Failing to dispose of the build contexts will leave the serverless process running indefinitely if there are build errors. This will cause issues with cicd platforms.

Perhaps ignore the context API altogether, since "rebuild" is used to manually trigger builds from esbuild's cache, and cache the results of "build" and call "build" again since serverless-esbuild is using its own caching mechanism.

Or moving the context disposal to another lifecycle hook might fix the serverless-offline issue.

kenstack-ltd (approved):

nathando1991:
Ok

floydspace (approved):
okay, I see there is a demand for this fix, gonna release it then.

Issue #445, opened 2023-03-08
Plugins do not seem to be recognised or run during build


danielwaghorn:
**Describe the bug**
We are compiling TypeScript based functions via the plugin, and our project uses TypeScript paths e.g. `@ourcompany/logger` to resolve libraries from a separate directory to be used within functions for common functionality.

In order to resolve these paths for bundling we are looking to use the [tsconfig-paths](https://www.npmjs.com/package/@esbuild-plugins/tsconfig-paths) plugin, but it appears that the plugins are never passed to esbuild / or executed.

We created a minimal plugin with a `console.log` in the `setup` method and it is never invoked. The `get plugins()` method within the project is definitely returning the plugins attribute from `custom.esbuild.plugins` and this resolves the file via `require` to return the `plugins` object as expected.

It appears however that `esbuild` never executes with the plugins. The `console.log` is never invoked.

**To Reproduce**
Steps to reproduce the behavior. If you have an example repository that would be even better:

The following minimal project was created to verify this behaviour.

_serverless.ts_
```typescript
const serverlessConfiguration: ServerlessAzureConfig = {
  service: 'my-api-dev',
  provider: {
    name: "azure",
    region,
    runtime: Runtime.NODE14,
    stage: 'dev',
  },

  functions: {
    health: {
      handler: "api/src/endpoints/health/healthGet.default",
      events: [
        {
          http: true,
          authLevel: "anonymous",
          name: "req",
          methods: ["GET"],
          route: "health",
        },
      ],
    },
  },

  plugins: [
    "serverless-dotenv-plugin",
    "serverless-esbuild",
    "serverless-azure-functions",
  ],

  custom: {
    esbuild: {
      plugins: "esbuild-plugins.js",
    },
  },
};

module.exports = serverlessConfiguration;
```

_esbuild-plugins.js_
```javascript
module.exports = [
  {
       name: "plugin:test",
       setup() {
            console.log('plugin was run');
       },
  },
];

```

_api/src/endpoints/health/healthGet.ts_
```typescript
export default const handler = async (): Promise<void> => {
  context.res = response(200, { health: true });
};
```

**Expected behavior**
We expect to see the output from the `console.log` in the terminal output, but it's never called or invoked. I suspect that the plugins are never presented to esbuild when it is invoked.

**Screenshots or Logs**
No logs relevant to this issue.

**Versions (please complete the following information):**

- OS: MacOS 13.1 (22C65)
[...]

PR #446, opened 2023-03-11
feat: adds support for `serverless-azure-functions` plugin

danielwaghorn:
Adds support for Azure Function Apps built in Node via the `serverless-azure-functions` plugin.

Currently when trying to use the plugin with the `azure` runtime it does not compile because there is not a function matcher available, only for AWS and GCP.

This commit adds a matcher for Azure, along with unit tests and supporting documentation in README.md.

## Context
I originally had #445 opened, but upon further investigation the `esbuild` binary was never being invoked because the plugin ignored all `azure` runtime functions. I have since created a matcher and can confirm that the `esbuild` plugins functionality is working correctly.

This PR is raised to increase the compatibility of the plugin with other serverless runtimes, but it might be pertinent to add a warning whenever a function is being skipped e.g. an unsupported runtime or cloud.

Temporarily published to https://www.npmjs.com/package/@danielwaghorn/serverless-esbuild for testing if you wish to install locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,9 +126,9 @@ The following `esbuild` options are automatically set.
 
 ## Supported Runtimes
 
-This plugin will automatically set the esbuild `target` for the following supported Serverless runtimes
+This plugin will automatically set the esbuild `target` for the following supported Serverless runtimes:
 
-AWS:
+###AWS 
 
 | Runtime      | Target   |
 | ------------ | -------- |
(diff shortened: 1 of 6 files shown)


floydspace (dismissed):
Hi @danielwaghorn, thanks for contribution, happy that it works with Azure as well

floydspace:
@danielwaghorn, but for some reason tests check failed, could you have a look ?

danielwaghorn:
Thanks @floydspace, I'll sort the e2e tests and update the PR. It looks like we'll need to split them out to test AWS and Azure scenarios separately.

floydspace:
@danielwaghorn make sense, perfectly each cloud must be covered, but feel free to just fix the tests, probably we need to use aws provider explicitly in examples

danielwaghorn:
Will do, I think testing that an Arm template is generated here is out of scope for the plugin.

danielwaghorn:
@floydspace just needed e2e snapshots updated, which I have done now.

floydspace (approved):

Issue #448, opened 2023-03-14
Add option for concurrency of zipping packages


fredrik:
**Is your feature request related to a problem? Please describe.**
For large projects with many functions, the packing stage can cause out-of-memory errors. This is due to the zipping of functions being done in parallel. (See https://github.com/floydspace/serverless-esbuild/blob/master/src/pack.ts#L174)

The problem is more likely to manifest with larger functions — the project I'm working ran into this issue when a new dependency caused the bundled function size to double.

**Describe the solution you'd like**
I'd like to limit the number of concurrent calls to `zip` in the package step using an optional configuration parameter.

**Describe alternatives you've considered**
Simply zipping one function (or a small pre-defined number of functions) at a time would also solve the problem, but might have other performance impacts.

PR #449, opened 2023-03-14
feat: zip concurrently with pMap

fredrik:
Here's a simple proof-of-concept PR which calls `zip` with pMap and a concurrency setting of 4. The previous behaviour was to zip all functions in parallel.

The background is described here: https://github.com/floydspace/serverless-esbuild/issues/448

Three potential improvements to this PR:

- If this is a fruitful path to go down, we will want to parametrise `concurrency`. We could use `buildOptions.concurrency` which is currently used to instruct esbuild's concurrency, but perhaps having its own "zip concurrency" option is better.

- The implementation can also be cleaned up. I have simply wrapped the zip logic in an inline function ("`zipper`") but this could be refactored somewhat nicer. There are references to variables defined in the outside scope which is perhaps okay but not ideal.

- I have added some ad-hoc logging which was useful to me but I am not sure what the project's logging policy is. 😎

### DIFF
--- a/src/pack.ts
+++ b/src/pack.ts
@@ -1,6 +1,7 @@
 import assert from 'assert';
 import path from 'path';
 
+import pMap from 'p-map';
 import fs from 'fs-extra';
 import globby from 'globby';
 import { intersection, isEmpty, lensProp, map, over, pipe, reject, replace, test, without } from 'ramda';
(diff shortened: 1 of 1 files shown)


floydspace:
hi @fredrik, could you please rebase on latest master, there were integration tests fixed recently ..

fredrik:
> hi @fredrik, could you please rebase on latest master, there were integration tests fixed recently ..

Rebased. Tests are green 🍏

floydspace (approved):
Thanks, @fredrik, look good to me. The GitHub diff sucks, shows too many changes

PR #451, opened 2023-03-20
fix: lint warnings

fredrik:
There were three warnings from eslint, two about console.log and one about BuildResult only used as a type.


floydspace (approved):
Fine to me, having messages as part of error

floydspace:
@fredrik I'm squashing, so no worries, sometimes I rewrite message

PR #466, opened 2023-05-23
Update README.md: Set target to 'node16'

yuyokk:
Hi folks,

Minor change to docs to set `target: 'node16'` instead of outdated node12.

Please let me know if this makes sense.
Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ The following `esbuild` options are automatically set.
 | `incremental` | N/A        | Cannot be overridden. Use `disableIncremental` to disable it           |
 | `outDir`      | N/A        | Cannot be overridden                                                   |
 | `platform`    | `'node'`   | Set `format` to `esm` to enable ESM support                            |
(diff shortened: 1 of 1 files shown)


floydspace:
Hi @yuyokk not sure what's the reasoning for the PR, but currently the `node12` is the default target https://github.com/floydspace/serverless-esbuild/blob/ebd4fd45df5c7b94880047c42c123f7d068d1c87/src/index.ts#LL269C9-L269C9

So just changing it in docs does not make sense, it'd be inconsistent with the current behaviour. However, I agree that it must be updated. So I will accept the PR if you make the real change of the default `target`, make sure tests are also green. Let me know if you need some help.

yuyokk:
@floydspace updated default value as well.

`npm run test` works locally. 

`npm run test:e2e` fails because of snapshot. Do I just update the snapshot and add to mr?

floydspace:
@yuyokk there is a PR checks which run e2e, so no worries.

floydspace (approved):

PR #469, opened 2023-06-01
feat: improves error message when handler is not found

alex-springer-sr:
When moving from a different bundler to serverless-esbuild it can be hard to find mistakes with the generic error message provided in the `extractFunctionEntries` method.
To give developers at least a hint where the issue can be found this change adds the function alias to the error message.


floydspace (approved):

floydspace:
Thanks @alex-springer-sr, any size of contribution is appreciated.

PR #474, opened 2023-07-07
feat: avoid adding unnecessary package.patterns property to functions

tvhees:
Serverless v3.32.0 added a warning when encountering `package.patterns` on a function if `service.package.individually` is not set to `true`:
https://github.com/serverless/serverless/releases/tag/v3.32.0

This warning is being emitted when esbuild is used to package a service because the plugin defaults to adding an empty array for `package.patterns` to every function. The core serverless code does not expect this property to be there unless specified in configuration and the plugin should match this behaviour.


floydspace (approved):
@tvhees thank you. looks good

Issue #477, opened 2023-07-26
Yarn packager excludes non-semver imports such as github references


flipscholtz:
**Describe the bug**

The yarn packager [assumes](https://github.com/floydspace/serverless-esbuild/blob/ef5b2d14f66efb64093b1e0997cb64aa52855647/src/packagers/yarn.ts#L26C2-L26C2) the format `<name>@<semver>` , where everything before the last `@` is the name, and after is the version.

This causes a problem for other formats such as this:

```
"dependencies": {
   "@metacask/kyc-token-client": "github:mcask/kyc-token-client#0.3.2"
}
```

Even if that is fixed and the package is included in the yarn packager results, it still looks like this:
`@metacask/kyc-token-client@git+ssh://git`.
This is because the github ssh URL contains an '@' which is unrelated to npm package naming conventions:
```
git+ssh://git@github.com/mcask/kyc-token-client.git#0.3.2
```
Splitting at the `@` and assuming the first bit is the package name is inaccurate in this case.
 
This means the package does not pass the depWhitelist comparison [here](https://github.com/floydspace/serverless-esbuild/blob/ef5b2d14f66efb64093b1e0997cb64aa52855647/src/pack.ts#L80C7-L80C7) and so is still excluded.

**To Reproduce**
Clone the minimal repo here: https://github.com/flipscholtz/serverless-esbuild-github-deps-test
1. Run `yarn install`
4. Note that `node_modules/@identity.com/package.json` imports a @metacask lib using a github reference:
```
 "@metacask/kyc-token-client": "git+ssh://git@github.com/mcask/kyc-token-client.git#0.3.2",
```
5. Run `yarn sls package`
6. Unzip the function `.serverless/test.zip`
7. Note that the packages node_modules has no `@metacask/kyc-token-client` dependency.

**Expected behavior**
The github-referenced dependency should be included in the packaged sls node_modules.

**Screenshots or Logs**
N/A

**Versions (please complete the following information):**

- OS: MacOS Ventura 13.0
- Serverless Framework Version: 3.33.0
- Plugin Version: 1.46.0

**Additional context**
Proposed solution:
1. If the package reference can't be parsed to valid semver, the packager shouldn't do the semver comparison and just include the package by default.
2. The package name returned by the packager should be tolerant of multiple `@` signs, such as in the case of github URL's, so it still returns the base name of the package accurately.
Maybe a regex like this to get the name: `^(@[^@]+|[^@]+)`

[...]

PR #478, opened 2023-07-26
feat: support non-semver versions in the yarn packager, for example 'github:' references

flipscholtz:
feat: support non-semver versions in the yarn packager, for example 'github:' references 

Issue [here](https://github.com/floydspace/serverless-esbuild/issues/477)

- Don't assume the format `<name>@<semver>` . Use a regex that also returns the correct package name for this format:
```
@org/package@git+ssh://git@github.com/org/package
```
(The last `@` in this case is part of the URL and does not denote the boundary between package name and version).

- If the version is not valid semver, don't exclude the package from the yarn packager results, as it may be a github reference for e.g. where semver doesn't always apply.

### DIFF
--- a/src/packagers/yarn.ts
+++ b/src/packagers/yarn.ts
@@ -1,5 +1,5 @@
 import { any, isEmpty, reduce, replace, split, startsWith } from 'ramda';
-import { satisfies } from 'semver';
+import { satisfies, valid as isValidSemver } from 'semver';
 
 import type { DependenciesResult, DependencyMap, PackagerOptions } from '../types';
 import { SpawnError, spawnProcess } from '../utils';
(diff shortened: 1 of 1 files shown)


floydspace:
Hi @flipscholtz . I run the tests for the feature, and they failed, could have a look
and is it applyed only for `yarn`? should it also be implemented for `npm` and `pnpm`?

flipscholtz:
> Hi @flipscholtz . I run the tests for the feature, and they failed, could have a look and is it applyed only for `yarn`? should it also be implemented for `npm` and `pnpm`?

Hi @floydspace 
I pushed a fix for the tests, they are passing for me now.
I am looking at npm and pnpm, will update asap.

Issue #158, opened 2021-07-03
serverless-esbuild should expose lifecycle events to be hookable


gabsn:
After looking for `lifecycleEvents` in the source code, I didn't find any hookable events I could use to customize the `serverless-esbuild` plugin behavior.

However it is common best practice to expose such events when you write a plugin (See [here](https://www.serverless.com/blog/advanced-plugin-development-extending-the-core-lifecycle)).

For example, `serverless-offline` exposes `init`, `ready` and `end` event that `serverless-esbuild` actually uses to build the module before launching serverless-offline (See [here](https://github.com/dherault/serverless-offline/blob/b39e8cf23592ad8bca568566e10c3db3469a951b/src/ServerlessOffline.js#L33)).

I would need this hook to copy some static assets before bundling the package with `esbuild`.

Let me know if you want me to open the PR.

floydspace:
Hi @gabsn, I have in mind to implement this similar way like `serverless-webpack` does, but cannot currently focus on it. I would appreciate your effort if you're willing to contribute.

gabsn:
Sorry finally I decided to go with `serverless-scriptable-plugin` and:
```
  scriptable:
    hooks:
      # serverless
      before:deploy:deploy: yarn build
      before:invoke:local:invoke: yarn build
      # serverless-offline
      before:offline:start: yarn dynamodb:start & yarn watch &
```

Then in my build command I can use esbuild with its native config.

Issue #108, opened 2021-04-07
Esbuild watch with sls offline not working


ipauler:
Hello.
I've added a watch pattern in esbuild config
```
watch:
      pattern: [ 'src/**/*.ts']
```
it's recompiling code after changes, but those changes are not working
like if i add `console.log('hello')` it will recompile but i won't see that console.log() anywhere
same for other changes

joelpickup:
Are you certain it's not this long standing issue with serverless-offline (https://github.com/dherault/serverless-offline/issues/864) ?

Enabling `--useChildProcesses` fixes this.

ipauler:
@joelpickup thank you will check 
looks like it

ipauler:
using childProcess flag helps, but after rebuild i have cold start which is not useful in my case 
anyway thanks, i'll be waiting for some fixes on serverless-offline side

olup:
I'll close this. Feel free to reopen if you need more help or find bugs on this lib's side.

henrypenny:
> --useChildProcesses

Detected unrecognized CLI options: "--useChildProcesses".

abhijeet-toptal:
try `--reloadHandler` with serverless offline

deviant32:
still not working for me with reloadHandler

naydichev:
also not working for me!

PR #488, opened 2023-09-15
Notify serverless-offline when watched files trigger rebuild

aldenquimby:
This is my first time contributing. Please let me know what you'd like to see in terms of tests, documentation, etc - thanks!

## Background
- With a default `serverless-offline` + `serverless-esbuild` setup, watching for changes does not work
- See https://github.com/floydspace/serverless-esbuild/issues/108

## Problem
- The `--reloadHandler` solution proposed in the above issue drastically hurts local dev performance on a large project

## Proposal
- Leverage https://github.com/dherault/serverless-offline/pull/1093 to tell `serverless-offline` when it needs to clear the cache

## Alternatives
- If https://github.com/floydspace/serverless-esbuild/issues/158 was built, this may not be necessary
- For example with `serverless-webpack`, which already supports lifecycle hooks, I use a custom plugin to solve this exact same problem:
```js
class OfflineInvalidate {
  constructor(serverless) {
    this.hooks = {
      'after:webpack:compile:watch:compile': () =>
        serverless.pluginManager.spawn('offline:functionsUpdated'),
    };
  }
}
module.exports = OfflineInvalidate;
```
- I'd still prefer this proposed PR, so that watching "just works" with a default setup

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -360,6 +360,7 @@ class EsbuildServerlessPlugin implements ServerlessPlugin {
     chokidar.watch(allPatterns, options).on('all', (eventName, srcPath) =>
       this.bundle(true)
         .then(() => this.updateFile(eventName, srcPath))
+        .then(() => this.notifyServerlessOffline())
         .then(() => this.log.verbose('Watching files for changes...'))
         .catch(() => this.log.error('Bundle error, waiting for a file change to reload...'))
     );
@@ -398,6 +399,10 @@ class EsbuildServerlessPlugin implements ServerlessPlugin {
         ...(patterns.length && { patterns }),
       };
     }
+  }
+
+  notifyServerlessOffline() {
+    this.serverless.pluginManager.spawn('offline:functionsUpdated');
   }
 
   async updateFile(op: string, filename: string) {


floydspace (approved):
I agree with your statement, currently, we don't have lifecycle events, so this workaround would be fair to have till the time we implement hooks, (if we ever do that)

pslm93:
Hi could you please provide a simple example for making hot reload work when utilizing serverless-offline + serverless-build? I am unable to get this to work. What should the watch pattern be? I see in examples that the pattern is an array but my IDE complains that the pattern must be a string. What should be done for serverless offline? 
I have the following in my serverless.yml file:
`
plugins:
  - serverless-domain-manager
  - serverless-esbuild
  - serverless-newrelic-lambda-layers
  - serverless-prune-plugin
  - serverless-offline
  
  custom:
     esbuild:
       external:
         - deasync
         - newrelic
       watch:
         pattern: "src/**/*.(js|ts)"
       bundle: true
       minify: true
       sourcemap: true
`

But when I have my application running by `sls offline`, when I update anything in my code, the changes don't trigger a rebuild. Thank you so much for your help.

aldenquimby:
@pslm93 any chance you can link to an example repo? I have a similar configuration, and it's working for me after this PR.

To debug, I think the first step is to determine if the problem is with esbuild, or with serverless-offline. Add keepOutputDirectory: true under esbuild in your serverless.yml so you can see the .esbuild directory. Change a source file, and you should see the JS output update accordingly automatically. If it doesn't update, the problem is with esbuild and your watch config. If it does update, the problem is with serverless-offline.

Issue #485, opened 2023-08-22
Externals packaging prioritises yarn.lock lookup


Dutzu:
**Describe the bug**
When packing externals, the Utils/findUpIO willl recursively search up the directory structure for a given file and the function findProjectRoot calls this function sequentially looking for yarn.lock, pnpm-lock.yaml and package-lock.json.

In case you are using pnpm or npm instead of yarn BUT you have a yarn.lock somewhere up the directory structure, it will cause a fail-ure.

**To Reproduce**

1. Setup a basic serverless with esbuild project, define any externals. Use npm with a package-lock.json file in your project
2. Somewhere, anywhere up the folder structure create a yarn.lock file.
3. Execute sls package.

**Expected behavior**
The build should work....

**Screenshots or Logs**

```JS
lifecycle:command:invoke:hook: [16] < before:package:createDeploymentArtifacts
Compiling to node18 bundle with esbuild...
Compiling with concurrency: Infinity
Compiling completed.
plugin:serverless-esbuild: Trying to create packager: npm
plugin:serverless-esbuild: Packager created: npm
lifecycle:command:invoke:hook: [16] > before:package:createDeploymentArtifacts
process: handle error
process: finalize { error:
   { Error: ENOENT: no such file or directory, open '/Users/anonymous/package.json'
       at Object.openSync (node:fs:600:3)
       at Object.readFileSync (node:fs:468:35)
       at readFileSync (/Users/anonymous/Projects/theProjectInQuestion/node_modules/serverless/lib/utils/fs/read-file-sync.js:7:24)
       at Utils.readFileSync (/Users/anonymous/Projects/theProjectInQuestion/node_modules/serverless/lib/classes/utils.js:78:12)
       at EsbuildServerlessPlugin.packExternalModules (/Users/anonymous/Projects/theProjectInQuestion/node_modules/serverless-esbuild/dist/pack-externals.js:212:51)
       at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
       at async before:package:createDeploymentArtifacts (/Users/anonymous/Projects/theProjectInQuestion/node_modules/serverless-esbuild/dist/index.js:78:17)
       at async PluginManager.runHooks (/Users/anonymous/Projects/theProjectInQuestion/node_modules/serverless/lib/classes/plugin-manager.js:530:9)
[...]

PR #502, opened 2023-10-09
fix: recursively search for packer lock file for all 3 packers supported

Dutzu:
This is a quick fix for issue #485.

I stumbled upon it by accident but basically I had a yarn.lock file in my home folder, and because the mechanism for recursively finding the lock file of the project runs up the directory chain outside of the scope of the repo, it could happen.

In the issue i described 2 possible solutions i could see, and decided to push this parallel search as an easy way to fix it. It should basically look recursively up the folder tree for either a package-lock, yarn-lock or pnpm-lock file and stop when it finds one of them rather than first looking for a yarn.lock up into infinity and if it can't find one, move on to the next possible packer lock file and so forth...

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -47,9 +47,9 @@ const isPathRoot = (p: string) => rootOf(p) === path.resolve(p);
 const findUpIO = (names: string[], directory = process.cwd()): IOO.IOOption<string> =>
   pipe(path.resolve(directory), (dir) =>
     pipe(
-      IO.of(names.map((name) => safeFileExistsIO(path.join(dir, name))).some((value) => value())),
-      IO.chain((exists: boolean) => {
-        if (exists) return IOO.some(dir);
+      IO.sequenceArray(names.map((name) => safeFileExistsIO(path.join(dir, name)))),
+      IO.chain((exist) => {
+        if (exist.some(Boolean)) return IOO.some(dir);
         if (isPathRoot(dir)) return IOO.none;
         return findUpIO(names, path.dirname(dir));
       })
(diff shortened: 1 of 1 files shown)


floydspace:
and I'm thinking should we reuse the https://github.com/sindresorhus/find-up
so we don't have to maintain this part

Dutzu:
Sorry about that. I thought I could quickly pull it off in the browser editor. FIxed the code now.

As for replacing it with find-up.....your call, but for such a small piece of logic, I wouldn't. Sounds like a pad-left type of scenario.

floydspace:
hey @Dutzu 
I'm ready to approve it, wanted to make a little fix, but couldn't in your branch, could you please either allow me to commit or simply make this patch
```patch
diff --git a/src/utils.ts b/src/utils.ts
index f1d59cd..a49442f 100644
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -47,9 +47,9 @@ const isPathRoot = (p: string) => rootOf(p) === path.resolve(p);
 const findUpIO = (names: string[], directory = process.cwd()): IOO.IOOption<string> =>
   pipe(path.resolve(directory), (dir) =>
     pipe(
-      IO.sequenceArray(names.map((name) => safeFileExistsIO(path.join(dir, name)))),
-      IO.chain((exist) => {
-        if (exist.some(Boolean)) return IOO.some(dir);
+      IO.of(names.map((name) => safeFileExistsIO(path.join(dir, name))).some((value) => value())),
+      IO.chain((exists: boolean) => {
```
if you are not familiar with functional programming, it's not the best practice to unwrap Monad (`... value() ...`) and then wrap it back (`IO.of`).
It's possible to do that within the Monad, `sequenceArray` lazily evaluates each file existence and transfers the aggregated result to the chained function

Dutzu:
Yep, you got it right, unfortunately I haven't done much functional programming. This was my first encounter with fp-ts.

I applied your patch and also gave you rights to the fork in case more touch-ups are needed.

floydspace (approved):
Thank you @Dutzu approved

PR #508, opened 2023-11-29
fix: remove the list of default excluded files

### LINKED ISSUE #315: Extra file (package.json) not included into final zips

cyberkrishna666:
In this PR:

- Removes `excludedFilesDefault` array and its usage in `pack.ts`

This will actually fix #315 and allow to include any of `package-lock.json`, `pnpm-lock.yaml`, `yarn.lock`, `package.json` in the final artifact using `patterns` in `serverless.yml`.
It won't change the previous behavior as those files are not included unless specified.

### DIFF
--- a/src/pack.ts
+++ b/src/pack.ts
@@ -38,8 +38,6 @@ function setFunctionArtifactPath(
   }
 }
 
-const excludedFilesDefault = ['package-lock.json', 'pnpm-lock.yaml', 'yarn.lock', 'package.json'];
-
 export const filterFilesForZipPackage = ({
   files,
   functionAlias,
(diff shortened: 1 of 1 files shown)


cyberkrishna666:
@floydspace please take a look 🙏

cyberkrishna666:
There's an issue with the `e2e:individually` test failing because `package.json` is included in each function packaged individually, regardless of whether `package.json` is in the `patterns` or not. The problem occurs when the test tries to unzip two archives into the same folder. Since `package.json` is in both archives, it asks for permission to overwrite the file.

Also, if we were to keep the default list of excluded files only for individual packaging, it causes issues when `package.json` is added to the global `patterns` in `serverless.yaml`. In this situation, `yarn` struggles to parse production dependencies due to `"serverless-esbuild": "file:../.."` in the `.esbuild` folder, causing an error.

A possible solution is to remove `devDependencies` from `package.json` after copying it to the `.esbuild` folder and to change the `unzip` command to only extract the handler's file. This approach should help fix both the file conflict and the dependency parsing problems.

PR #510, opened 2023-12-01
feat: add support for resolve extensions esbuild setting

JohnGrisham:
Support the ability to add custom resolution extensions via ESbuild settings. This is a feature of ESbuild and will allow for customization of lambda entrypoints. My team will be using this to support building different files based on the target environment we are deploying to.

Example:

`myLambda.dev.ts`
`myLambda.prod.ts`
`myLambda.ts`

When the environment we build for matches the extension of the file then that entrypoint will be used otherwise it will fallback to the default file resolutions in the order of our choosing. If the user of this plugin doesn't pass a `resolveExtensions` ESbuild setting then it will fallback to the array that was already defined by the repo.

https://esbuild.github.io/api/#resolve-extensions

## Notes:

When using custom extensions as the entrypoint for lambdas AWS can complain if the file uses a custom extension. To resolve this I have also added a `stripEntryResolveExtensions` option to the plugin which will remove custom extension prefixes if they are used in the resolveExtensions list. This will result in files that have extensions like `test.custom.js` to be renamed as `test.js` in the eventual packed file. I'm open to suggestions if there is a better way to handle this or if this behavior should be the default for any custom extensions assuming the approach in this PR is approved.


JohnGrisham:
@floydspace I have made the requested changes, please let me know if there is anything else I can clear up or change.

floydspace (approved):

floydspace:
LGTM, I gonna merge it, hope it will be useful for others

thank you for contribution

Issue #514, opened 2023-12-13
Esbuild cannot rebuild error


0gebey:
We built a custom plugin to invoke our lambdas locally with local RabbitMQ messages and with the latest version(1.50), we are facing with "Esbuild cannot rebuild" error when we try to invoke local lambdas programmatically. Every 3rd trial throws the same error. It is fine with cli command "serverless invoke local --function functionName".

You can reproduce the error by trying to invoke your lambda like that;

```
// Invoke the function
            const AwsInvoke = this.serverless.pluginManager.plugins.find(
              (plugin) => plugin.constructor.name === "AwsInvokeLocal"
            );

            AwsInvoke.serverless.pluginManager
              .spawn("invoke:local", { function: functionName })
              .catch((error) => {
                console.log("error:", error);
                throw new Error(
                  `Error invoking local AWS Lambda function ${functionName}: ${
                    error as string
                  }`
                );
              });
```

- OS: Mac
- Serverless Framework Version: 3.35.2
- Plugin Version: 1.50.0 (We realized that the issue starts with 1.48.1)

We figured out, "disposeContext" function which has added with 1.48.1 causes the problem but we would like to learn about the actual reason of it.

PR #515, opened 2023-12-13
Implement 'disposeContext' option in Configuration and EsbuildFunctionDefinitionHandler

0gebey:
**Description:**

This pull request addresses the issue explained in [#514](https://github.com/floydspace/serverless-esbuild/issues/514). We identified a challenge regarding the handling of context disposal in the serverless-esbuild plugin, which caused error and changed behavior of Lambda functions on every 3rd local invoke.

**Changes Made:**

**New _'disposeContext'_ Property:** I introduced a new property named **disposeContext** which has a default value as **'true'** in both **Configuration** and **EsbuildFunctionDefinitionHandler** interfaces. This property allows users to control whether the _disposeContexts()_ function should dispose of the context.

**Configuration Flexibility:** The **disposeContext** option can be set globally within the esbuild configuration in the serverless configuration file. This global setting applies to all Lambda functions utilizing the specific configuration.
Function-Level Override: Individual functions can override the global setting by including the **disposeContext** property in their own configurations.

**Backward Compatibility:** Careful consideration has been given to ensure backward compatibility with these changes. For existing users of the **serverless-esbuild** library who have not specified the **disposeContext** option in their configurations, there will be no disruption in functionality. This is because the **disposeContext** property is set to a default value of **'true'**. Consequently, the context disposal behavior will continue to operate as it did previously, maintaining the expected behavior for current users. This approach ensures a seamless transition for users updating to the version of the library that includes these new changes.

**Testing:** The testing for these changes was conducted within our own development environment, specifically tailored to our use case of programmatically invoking Lambda functions locally.

**Impact of Changes:**

This enhancement provides greater control over context management in Lambda functions, potentially improving performance and resource management for users of the serverless-esbuild plugin. It offers both a global and a function-specific configuration approach, catering to diverse use cases and deployment strategies.

I believe this feature can solve the problem of the ones who were facing with the same error and am looking forward to feedback from the maintainers and community.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ See [example folder](examples) for some example configurations.
 | `skipBuild`            | Avoid rebuilding lambda artifacts in favor of reusing previous build artifacts.                                                                                                                    | `false`                                             |
 | `skipBuildExcludeFns` | An array of lambda names that will always be rebuilt if `skipBuild` is set to `true` and bundling individually. This is helpful for dynamically generated functions like serverless-plugin-warmup. | `[]`                                                 |
 | `stripEntryResolveExtensions` | A boolean that determines if entrypoints using custom file extensions provided in the `resolveExtensions` ESbuild setting should be stripped of their custom extension upon packing the final bundle for that file. Example: `myLambda.custom.ts` would result in `myLambda.js` instead of `myLambda.custom.js`.
-
+| `disposeContext` | An option to disable the disposal of the context.(Functions can override the global `disposeContext` configuration by specifying their own `disposeContext` option in their individual configurations.) | `true`
 #### Default Esbuild Options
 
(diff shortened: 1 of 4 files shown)


floydspace:
hi @0gebey thank you for the detailed explanation of the issue you are facing, and the PR. 
could you please also update the readme with a little explanation and default value, as you are adding a new configuration option.

floydspace:
and could you also check tests

floydspace:
@0gebey could you please also rebase your branch, I fixed the recent e2e tests issue, blocking PRs

0gebey:
> @0gebey could you please also rebase your branch, I fixed the recent e2e tests issue, blocking PRs

Can you check it again ? I believe it should be okey right now.

floydspace (approved):

Issue #536, opened 2024-05-20
update esbuild peerDependency (0.20 -> 0.21)


naotama2002:
esbuild 0.21 is out, need to update the peerDependency.

lynxSven:
esbuild 0.22 is now out

rion18:
esbuild 0.23 is now out :thinking:

rion18:
oh my. serverless-esbuild gets updated and esbuild 0.24.x gets released :(

floydspace:
I'm working on it , that was indeed a face palm moment to me

PR #537, opened 2024-05-20
feat: #536 update esbuild version to 0.21.0

naotama2002:
- update esbuild

Attempts a solution for https://github.com/floydspace/serverless-esbuild/issues/536

### DIFF
--- a/examples/complete/package.json
+++ b/examples/complete/package.json
@@ -9,7 +9,7 @@
   },
   "devDependencies": {
     "@types/node": "^18.7.21",
-    "esbuild": "^0.20.0",
+    "esbuild": "^0.21.0",
     "serverless": "^3.22.0",
     "serverless-esbuild": "file:../..",
     "serverless-offline": "^10.2.1",
--- a/examples/config/package.json
+++ b/examples/config/package.json
@@ -9,7 +9,7 @@
   "devDependencies": {
     "@types/lodash": "4.14.185",
     "@types/node": "^18.7.21",
-    "esbuild": "^0.20.0",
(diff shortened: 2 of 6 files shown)


johnsonps08:
It seems as though @aws-sdk/client-cloudformation is no longer compatible with Node <16. Do the checks need to be updated to remove 14 and this be declared a breaking change? Or is there some other path for this to proceed?

PizzaPete (approved):

naotama2002:
@johnsonps08 

Thanks for confirming.

First, the fact that the test fails with Node 14 has nothing to do with the esbuild version update.

The cause is the end of support for Node.js 14.x in aws-sdk-js-3 [issue #6034](https://github.com/aws/aws-sdk-js-v3/pull/6034). 
Therefore, since the release of [aws-sdk-js-3 v3.567.0](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.567.0) which includes #6034, tests for Node.js 14 will fail.

> Do the checks need to be updated to remove 14 and this be declared a breaking change?

Since Node.js 14 is not listed in the supported runtimes, I think supporting Node.js 16 and 18 is sufficient.
- https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

Should we create an issue for end support for Node.js 14.x?

naotama2002:
@johnsonps08 
Thanks. issue and PR created.

> Should we create an issue for end support for Node.js 14.x?

issue: https://github.com/floydspace/serverless-esbuild/issues/543
PR: https://github.com/floydspace/serverless-esbuild/pull/544

Issue #388, opened 2022-11-07
esbuild Process Using 13+ GB of Memory


cdalton713:
**Describe the bug**
The esbuild process is using a ridiculous amount of memory.  Our backend stack is relatively small, less than two dozen endpoints.  

**To Reproduce**

Serverless Config: 

```
custom: {
...
    esbuild: {
      bundle: true,
      minify: false,
      sourcemap: true,
      incremental: true,
      exclude: ["aws-sdk", "pg-hstore"],
      target: "node16",
      define: { "require.resolve": undefined },
      platform: "node",
      concurrency: 5,
      // plugins: "./esbuild-plugin.js",
      packager: "pnpm",
      packagePath: "./package.json",
    },
    ...
}
   ```

**Expected behavior**
Not 15 GB worth of memory for a build.

**Screenshots or Logs**
![Uploading CleanShot 2022-11-07 at 10.36.28@2x.png…]()

**Versions (please complete the following information):**

- OS:  M1 Macbook 12.2.1
- Serverless Framework Version: 3.24.0
- Plugin Version: 0.15.10

Hideman85:
I'm having the same issue, I try to run esbuild in CICD with limited ram capacity and the process is just growing exponentially to death of course. Can we force esbuild to run under a certain amount of ram for production build?

cdalton713:
Setting `concurrency = 1` and `disableIncremental = true` has worked as an interim fix; this reduced esbuid's memory usage to around 600 mb for us.  However, these settings of course affect build times - local development is now becoming increasingly difficult because each rebuild take so long to complete...

Hideman85:
Oh thank you so much men, it saved me a lot of pain :+1: and actually in my side the speed is not that much impacted.

mhamadyhya1:
Thanks very much for the solution

zack37:
I'm having this same problem, even after updating to latest. I've set concurrency to 1 but still have really high cpu and memory usage

esbuild.config.js
```js
module.exports = () => ({
  // concurrency: os.availableParallelism(),
  concurrency: 2,
  target: ["node18"],
  sourcemap: true,
  sourcesContent: false,
  platform: "node",
  format: "cjs",
  loader: loaderMap,
  exclude: ["*"],
  packager: "yarn",
  keepNames: true,
  packagerOptions: {
    ignoreLockfile: true
  },
  watch: {
    pattern: ["./**/*.(js|ts|graphql|html)"],
[...]

PR #540, opened 2024-05-26
Fix memory leak when skipRebuild is true

alexwillingham:
This PR addresses https://github.com/floydspace/serverless-esbuild/issues/388

It is inspired by https://github.com/floydspace/serverless-esbuild/issues/388#issuecomment-1773183932

I run several different services locally in parallel with serverless-offline and serverless-esbuild. The largest one of those was chewing up 30+ GB of RAM just to build and run locally, and it didn't come down after the build was done.  We kept having to bump our GitHub Actions to larger runners in order to be able to run our integration tests chiefly because of this memory leak.

We have been using this fork for a month; The memory usage is far less and there are no apparent issues. By using the fork, we've cut our GitHub Action spend in half.

### DIFF
--- a/src/bundle.ts
+++ b/src/bundle.ts
@@ -1,5 +1,6 @@
 import assert from 'assert';
 import type { BuildOptions } from 'esbuild';
+import * as pkg from 'esbuild';
 import fs from 'fs-extra';
 import pMap from 'p-map';
 import path from 'path';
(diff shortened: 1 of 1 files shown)


alexwillingham:
Notably, I took the approach of fixing this only for the case where `skipRebuild` is true because that seemed more easily achievable. The same memory leak may exist for the case where `skipRebuild` is false and could potentially be addressed later.

alexwillingham:
Heyo, just let me know if anything is needed on my part to get an approval here. Happy to help.

aviadsteps (approved):

floydspace (approved):

floydspace:
looks safe to merge, sorry for ignoring it for so long time

Issue #543, opened 2024-06-18
feat: end support for Node.js 14.x


naotama2002:
**Is your feature request related to a problem? Please describe.**
Node.js tests are failing because aws-sdk-js-3 (@aws-sdk/client-cloudformation), on which examples/minimal/ depends, no longer supports Node.js 14
ex. https://github.com/floydspace/serverless-esbuild/pull/537/checks

The cause is the end of support for Node.js 14.x in aws-sdk-js-3 [issue #6034](https://github.com/aws/aws-sdk-js-v3/pull/6034). 
Therefore, since the release of [aws-sdk-js-3 v3.567.0](https://github.com/aws/aws-sdk-js-v3/releases/tag/v3.567.0) which includes #6034, tests for Node.js 14 will fail.

**Describe the solution you'd like**
End support for Node.js 14.x.

**Describe alternatives you've considered**
"I have no ideas.

**Additional context**
Since Node.js 14 is not listed in the supported runtimes, I think supporting Node.js 16 and 18 is sufficient.
- https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

PR #544, opened 2024-06-18
feat: end support for Node.js 14.x

naotama2002:
End support for Node.js 14.x.
- Attempts a solution for https://github.com/floydspace/serverless-esbuild/issues/543


letharion:
Looks good, is there anything preventing this from being merged? If so I'd like to try to help.

PR #553, opened 2024-09-24
upgrade esbuild to 0.24

floydspace:
+ migrate utils from fp-ts to effect


package.json
```diff
@@ -92,11 +94,11 @@
     "mock-spawn": "^0.2.6",
     "prettier": "^2.8.1",
     "semantic-release": "^19.0.5",
-    "ts-jest": "^29.0.3",
-    "typescript": "^4.9.4"
+    "ts-jest": "^29.2.5",
+    "typescript": "~5.5.4"
   },
   "peerDependencies": {
-    "esbuild": ">=0.8 <0.24",
```
ebisbe, line 99: Is there any reason to be limited to 0.24? The 0.25 is already up and I rather have a more loose peerDependecy

ebisbe: Thanks for the explanation. I ( wrongly ) assumed that they were using semversionig. I'll try to make a PR.

floydspace: I believe they are using semver, but semver for v0 means: minor increment can be a breaking change

PR #555, opened 2024-09-26
Add dependency on @effect/schema

letharion:


letharion:
Already done.

floydspace:
ah I noticed your fork, was not paitient to wait a bit, so thanks anyway

Issue #557, opened 2024-10-10
Since `1.54.1` the package bundling is no longer "deterministic", that means we are re-deploying packages with no code changes.


bnegrao:
**Describe the bug**
The [commit](https://github.com/floydspace/serverless-esbuild/commit/162d9bc7f0e87d63cd93adb0f4756d01769da79d) in version 1.54.1 introduced a bug in the bundling process. The produced zip files are always binary different, with no code changes involved.

The consequence is that the packages are being re-deployed every time, as if they contain changes.

**To Reproduce**
In any repository, run:
```bash
# create a package
npx sls package
# save the zip file
mv .serverless/*.zip first.zip
# repeat package creation
npx sls package
# save the second zip file
mv .serverless/*.zip second.zip
# Now compare them
diff first.zip second.zip
# diff will tell the "binary files differ".
```
The cause is the modification date. You can `unzip -l first.zip` and `unzip -l second.zip` to see.

**Expected behavior**
The zip files should be bundled in a "deterministic" way, that means, bundling the same code should always produce the exact same bundle.

**Additional context**
If you downgrade to version 1.53.0, the plugin works in the expected way.

bnegrao:
I created a PR to fix it: https://github.com/floydspace/serverless-esbuild/pull/558

PR #558, opened 2024-10-12
Revert "fix: preserve mtime when zipping with the node zipper (#539)"

bnegrao:
This PR reverts commit 162d9bc7f0e87d63cd93adb0f4756d01769da79d, which introduced non-deterministic behavior in the bundling process, leading to unnecessary re-deployments even when no code changes occurred. This change addresses issue [#557](https://github.com/floydspace/serverless-esbuild/issues/557).

### DIFF
--- a/src/tests/util.test.ts
+++ b/src/tests/util.test.ts
@@ -18,29 +18,15 @@ describe('utils/findProjectRoot', () => {
 });
 
 describe('utils/zip', () => {
-  const mtime = new Date(2024, 0, 1, 0, 0, 0, 0);
-
   beforeEach(() => {
     mockFs({
-      '/src': mockFs.directory({
-        mtime,
-        items: {
-          'test.txt': mockFs.file({
-            mtime,
-            content: 'lorem ipsum',
-          }),
-          modules: mockFs.directory({
-            mtime,
-            items: {
-              'module.txt': mockFs.file({
-                mtime,
-                content: 'lorem ipsum 2',
-              }),
-            },
-          }),
+      '/src': {
+        'test.txt': 'lorem ipsum',
+        modules: {
+          'module.txt': 'lorem ipsum 2',
         },
-      }),
-      '/dist': mockFs.directory({ mtime }),
+      },
+      '/dist': {},
     });
   });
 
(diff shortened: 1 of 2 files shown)


floydspace:
hi @bnegrao , thanks, I think it's fair to revert it, @iainlane we might need to have a better solution for this, like having configurtation option for zipper, for now I just let it revert

floydspace (approved):

iainlane:
Any idea what the problem is? Is it to do with the target system or something?

iainlane:
A testcase to reproduce would be nice; the one which is being reverted here is asserting that the results _are_ deterministic.

iainlane:
Actually, I suspect what is happening is that generated files are updated on a re-package. Isn't that what is expected? The zip _should_ be different if the input files have been updated (touched).

So - I think reverting this was the wrong move. The old behaviour wasn't really correct, and the new one was. If you're repackaging and that involves updating the generated code, you should indeed expect a different checksum on your zip.

In other words:

> The zip files should be bundled in a "deterministic" way, that means, bundling the same code should always produce the exact same bundle.

Disagree that we are seeing "nondeterminism".

iainlane:
I suspect if we took a look at the `ls -lR` of the zips we'd see the (generated) files have different mtimes.

Unless I understand [this](https://github.com/bnegrao/serverless-esbuild/blob/17cc8e6365f82accc541fcfc59cd2096edd862c2/src/utils.ts#L109
) wrong (could be), we're taking the mtime of the source file itself - the file which is being added to the zip a few lines later. (On the next line we skip directories)

floydspace:
> Isn't that what is expected?
this is what I expect: no matter what changes are in the source code, if bundling produces the same bundle from content point of view, it should not trigger cloudformation changeset

iainlane:
Well if that's the position then you disagree with my initial change as far as I can see. When are people doing this kind of no-change repackage?

If you ask me I'd state it as this: if you run a no-op rebuild (why?) then you should expect to get a redeploy, because your files did change - even if just in modification time.

Which is a conflicting position to yours, and unless you find that convincing, as the maintainer your view should be the one to prevail.

It's no problem. I'll go and look for another way to deploy in which I can have working bundled static assets. :-)

PR #560, opened 2024-11-22
feat: add node 22 support for AWS

robertbeal:
Node 22 runtime (current LTS) was recently added for Lambda:
https://aws.amazon.com/blogs/compute/node-js-22-runtime-now-available-in-aws-lambda/


floydspace:
thank you @robertbeal

PR #567, opened 2025-02-21
chore: update esbuild compatibility to include 0.25

strmer15:
The 0.25.0 version of `esbuild` was [released a couple of weeks ago](https://github.com/evanw/esbuild/releases/tag/v0.25.0) - it fixes various issues, but also restricts access to the esbuild dev server to fix a security vulnerability.

This change updates the peer dependency compatibility range and also the dev dependency version so that users of `serverless-esbuild` can upgrade to `esbuild` 0.25.0 without getting peer dependency warnings from their package manager.

### DIFF
--- a/src/tests/index.test.ts
+++ b/src/tests/index.test.ts
@@ -189,8 +189,8 @@ describe('Move Artifacts', () => {
     });
 
     it('should skip function if skipEsbuild is set to true', async () => {
-      // @ts-ignore
-      fs.existsSync.mockReturnValue(true);
+      jest.mocked(fs.existsSync).mockReturnValue(true);
+
       const hello3 = { handler: 'hello3.handler', events: [], skipEsbuild: true };
       const plugin = new EsbuildServerlessPlugin(
         mockServerlessConfig({
--- a/src/types.ts
+++ b/src/types.ts
@@ -129,7 +129,7 @@ export interface BuildContext {
 /** Documentation: https://esbuild.github.io/api/#serve-return-values */
 interface ServeResult {
   port: number;
-  host: string;
+  hosts: string[];
 }
 
 export type JSONObject = any;
--- a/.github/workflows/pull.yml
+++ b/.github/workflows/pull.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [18, 20, 22]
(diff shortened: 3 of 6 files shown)


floydspace:
Hi @strmer15, thanks for the PR.
that is not so simple to just update the version, we also need make sure it works :)
image
can you please check failing tests and also do functional test? otherwise I can pick it up later in `undefined` time

floydspace:
FYI @ebisbe :)

src/tests/index.test.ts
```diff
@@ -189,8 +189,8 @@ describe('Move Artifacts', () => {
     });
 
     it('should skip function if skipEsbuild is set to true', async () => {
-      // @ts-ignore
```
strmer15, line 192: This was giving a warning whenever running tests; the fix is to use `jest.mocked` to wrap the function to get the mock types to work correctly.

src/types.ts
```diff
@@ -129,7 +129,7 @@ export interface BuildContext {
 /** Documentation: https://esbuild.github.io/api/#serve-return-values */
 interface ServeResult {
   port: number;
-  host: string;
+  hosts: string[];
```
strmer15, line 132: This seems to have changed in the latest `esbuild`, [the docs say](https://esbuild.github.io/api/#serve-return-values) it should be a string array called `hosts` now.

strmer15:
@floydspace the node 22 e2e tests are failing because Node 22.12+ now turns on `require` module syntax detection by default - because one of the `serverless` entry points uses a top-level await, it then blows up with an async module error.

See [this mention ](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V22.md#requireesm-is-now-enabled-by-default) of the new behavior in the Node 22 release notes. Also, there's [this comment](https://github.com/serverless/serverless/issues/12936#issuecomment-2576970783) in a Node 22 issue on the `serverless` repo.

I could get around it by using `npx --node-options='--no-experimental-require-module' sls package` in the `Makefile`, but I'm not sure if that's what you'd like to do to fix it. This didn't start with this PR (I verified locally that Node 22 fails on `master`), but I think the test matrix is now hitting it because the last CI run was before Node 22.12 got released (and it's now getting Node 22.13.1).

floydspace:
@strmer15 I will look at it, thanks for checking

floydspace (approved):

PR #574, opened 2025-06-26
chore: switch to tinyglobby

benmccann:
https://npmgraph.js.org/?q=globby - 23 dependencies
https://npmgraph.js.org/?q=tinyglobby - 2 dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,10 +21,10 @@
         "effect": "^3.8.3",
         "execa": "^5.1.1",
         "fs-extra": "^11.1.0",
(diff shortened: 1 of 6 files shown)


floydspace:
hi @benmccann 
thanks for contribution, could you pls explain what motivated you to make this fix
and pls apply prettier formater changes properly

floydspace:
I don't think this switch will solve the issues you are describing, because `globby` is already a dependency of the `serverless` package itself, which anyway has to be installed by end user.
so, in fact, total, with this change end user will have + one more + 2 dependencies.

Issue #585, opened 2026-05-20
Concurrent builds collision when using native zip


artema:
**Describe the bug**

When native zip option is enabled, and multiple Serverless projects are performing builds at the same time, there is a possibility that `serverless-esbuild` will use the same temporary directory for multiple tasks, resulting in wrong code being packaged.

**To Reproduce**

1. Set nativeZip to `true`
2. Run `sls package` in parallel on multiple projects

**Expected behavior**
Each process is using its own temporary directory. Instead, code from another process is packaged.

**Versions (please complete the following information):**

- OS: Mac
- Serverless Framework Version: 3.38.0
- Plugin Version: 1.57.1

PR #586, opened 2026-05-20
fix: ensure unique temp directory during zip

artema:
Fixes temp path collisions when running concurrent builds #585


floydspace (approved):
@artema Spasibo bratan for contribution, that makes sense

